PR #2, opened 2015-04-09
Incorporating a feedback from KostyaSha and downgraded core to 1.545

kohsuke:
See
https://github.com/jenkinsci/docker-commons-plugin/commit/2e03b78e1c3ec3407b0869e7dfa4416dd77b7d02#commitcomment-10646991

I don't think 1.580 is "latest core", but I have no objections to rely
on 1.545 if he needs it, hence the change.

This version of the core apparently relied on older apache commons-io,
so Charset.UTF8 had to be taken from somewhere else.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>1.580.1</version>
+    <version>1.545</version>
     <relativePath/>
   </parent>
   <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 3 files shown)


kohsuke:
@reviewbybees

recampbell:
:+1:

alecharp:
:+1:

jglick:
Why 1.545? Do you mean 1.554?

jglick:
(so :-1: until that is explained)

jglick:
Also beware that 1.554 cannot be run on Java 8, complicating development of plugins using that baseline. You need 1.554.1 or later.

KostyaSha:
I don't need ancient cores, but i want minimise future PRs that merged and exists in plugins without releases. I was forced to review naginator commits and check why version was increased instead of CB person that sent decrease PR. Better define and document now instead of spending time later.

kohsuke:
Amended the commit with 1.554.1 and pushed the change.

PR #3, opened 2015-04-09
[WiP] Sketching a DockerClient API

tfennelly:
WiP atm.

Stab at a `DockerClient` that executes `DockerCommand`s and uses `KeyMaterial` etc.

Functionality from Oki Docki plugin `Docker` class:

- [x] hasImage (via `DockerInspectCommand.exists(...)`)
- [x] pullImage (via `DockerPullCommand`)
- [x] buildImage (via `DockerBuildCommand`).
- [ ] `DockerBuildCommand` to provide access to image fingerprint?
- [ ] Add test for `DockerBuildCommand`.
- [x] kill (via `DockerKillCommand`)
- [x] rm (via `DockerRMCommand`)
- [x] runDetached (via `DockerRunCommand`). Populates `ContainerRecord`. See `DockerRunCommandTest`
- [x] `DockerInspectCommand`. Used by `DockerRunCommand` to get the container info for the `ContainerRecord`. Misc other uses too.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,6 +57,12 @@
       <version>1.3</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.main</groupId>
+      <artifactId>jenkins-core</artifactId>
+      <version>1.580.1</version>
+      <scope>provided</scope>
+    </dependency>
   </dependencies>
 
 </project>
--- a/pom.xml
+++ b/pom.xml
@@ -57,12 +57,6 @@
       <version>1.3</version>
       <scope>provided</scope>
     </dependency>
-    <dependency>
-      <groupId>org.jenkins-ci.main</groupId>
-      <artifactId>jenkins-core</artifactId>
-      <version>1.580.1</version>
-      <scope>provided</scope>
-    </dependency>
   </dependencies>
 
 </project>
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerClient.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerClient.java
@@ -47,10 +47,6 @@ public class DockerClient {
         this.launcher = launcher;
     }
 
-    public Launcher getLauncher() {
-        return launcher;
-    }
-
     public DockerClient setKeyMaterial(KeyMaterial keyMaterial) {
         this.keyMaterial = keyMaterial;
         return this;
@@ -86,5 +82,20 @@ public class DockerClient {
             dockerCommand.setErr(err.toString());
         }
     }
-    
+
+    /**
+     * Who is executing this {@link DockerClient} instance.
+     *  
+     * @return a {@link String} containing the <strong>uid:gid</strong>.
+     */
+    public String whoAmI() throws IOException, InterruptedException {
+        ByteArrayOutputStream userId = new ByteArrayOutputStream();
+        launcher.launch().cmds("id", "-u").stdout(userId).quiet(true).join();
+
+        ByteArrayOutputStream groupId = new ByteArrayOutputStream();
+        launcher.launch().cmds("id", "-g").stdout(groupId).quiet(true).join();
+        
+        return String.format("%s:%s", userId.toString().trim(), groupId.toString().trim());
+
+    }    
 }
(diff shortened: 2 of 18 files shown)


tfennelly:
@reviewbybees please, especially @jglick

pom.xml
```diff
@@ -57,6 +57,12 @@
       <version>1.3</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.main</groupId>
+      <artifactId>jenkins-core</artifactId>
+      <version>1.580.1</version>
+      <scope>provided</scope>
+    </dependency>
```
jglick: Huh??

src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerClient.java
```diff
@@ -0,0 +1,101 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2014, CloudBees, Inc.
```
KostyaSha: 2015 !

src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerInspectCommand.java
```diff
@@ -0,0 +1,114 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2014, CloudBees, Inc.
```
KostyaSha: 2013 for new file? and 2015 is now

jglick:
```
…/src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryToken.java:[45,54] error: <anonymous org.jenkinsci.plugins.docker.commons.DockerRegistryToken$1> is not abstract and does not override abstract method checkRoles(RoleChecker) in RoleSensitive
```

possibly because of the weird `jenkins-core` dep which should be deleted.

pom.xml
tfennelly: So we shouldn't use `Launcher` is what you're saying?

KostyaSha:
CloudBees are inventing another one docker api library? Really? Why not use https://github.com/docker-java/docker-java/tree/master/src/main/java/com/github/dockerjava/api/command ??

tfennelly:
@KostyaSha well that's not out of the question yet ... what this one is doing is using the Jenkins `Launcher`, deal with fingerprints etc. It's really coming from the Oki Docki plugin. But yeah, might change it to use an external lib if it looks like that might be better (spotify also have one ... no idea what it's like).

jglick:
This is certainly not an attempt to make a general-purpose Docker API library. Only to provide convenient ways of integrating with other things already in this plugin, namely credentials and fingerprints.

As to whether using the CLI or a Java library is the better choice, I think this is an open question, but I tend to find the CLI a more reliable option. It also allows use from long-running builds such as Workflow: some Docker commands can run for a very long time (for example if they need to pull base layers), and you would prefer these things to run uninterrupted even if the slave agent goes away. Whether that mode is possible using the current code proposed here is another question: would probably need to make use of `durable-task`.

KostyaSha:
@jglick afaik library has interface and now only java implementation exists, it seems logical to extend existed library to work with binary if you really want deal with all this wrappers and their compatibilities

KostyaSha:
https://github.com/docker-java/docker-java/blob/master/src/main/java/com/github/dockerjava/api/command/DockerCmdExecFactory.java ?

KostyaSha:
Also, as you mentioned in IRC having weird implementations in "common" plugin is not acceptable.

jglick:
> having weird implementations in "common" plugin is not acceptable

I was discussing extensions/descriptors.

jglick:
> it seems logical to extend exist[ing] library to work with binary

Perhaps. That sounds like a big task, not necessarily welcome by the maintainers of that library.

src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerClient.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    public int launch(@Nonnull DockerCommand dockerCommand) throws IOException, InterruptedException {        
+        EnvVars envVars = new EnvVars();
+        
+        // Let the impl perform any impl specific setup.
+        dockerCommand.preLaunch();
+
+        if (keyMaterial != null) {
+            envVars.putAll(keyMaterial.env());
```
jglick: When do you call `close`?

tfennelly:
Closing this after talking to @jglick.

PR #4, opened 2015-04-09
Remove DockerRegistryCredentials class that pollutes Credentials type space

stephenc:
@reviewbybees


jglick:
:+1:

Do you have a separate plan for `DockerServerCredentials`?

stephenc:
DockerServerCredentials is "ok".

I still maintain a regular CertificateCredentials would suffice, with
perhaps a marker interfve to expose the issuing CA certificate, but we can
always pull that back out later. It's the pointless class-envy of
DockerRegistryCredential that I want to avoid... If we want to "colour"
credentials the way the credentials API is designed is to handle that via
domain spec/req pairs and credential functionality matching... Otherwise
how do users know "which user/pass credential type to create"

On Thursday, April 9, 2015, Jesse Glick <notifications@github.com> wrote:

> [image: :+1:]
>
> Do you have a separate plan for DockerServerCredentials?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/docker-commons-plugin/pull/4#issuecomment-91330464>
> .
>


-- 
Sent from my phone

jglick:
> `DockerServerCredentials` is "ok".

Except that (as I noted in `ConfigTest`) there are no implementations! So there is no way for the user to create any credentials, nor any way for us to test that they would work if created. Something is missing here.

stephenc:
On Friday, April 10, 2015, Jesse Glick <notifications@github.com> wrote:

> DockerServerCredentials is "ok".
>
> Except that (as I noted in ConfigTest) there are no implementations! So
> there is no way for the user to create any credentials, nor any way for us
> to test that they would work if created. Something is missing here.
>

This is a lesser issue in the design to my mind, so for now I am willing to
let it pass ;-)

> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/docker-commons-plugin/pull/4#issuecomment-91535426>
> .
>


-- 
Sent from my phone

kohsuke:
I'm sorry, but this is too much. I just can't take it any more.

We had 1hr+ of painful session where I let you have pretty much everything you wanted. We talked about this one too, you asked if I wanted you to write it, and this is the only thing I declined. And next morning you do it anyway and ask me to take it.

You can't pick every fight like that. This has left the realm of creating something better together long time ago. This is you speaking over me long and loud until I'm so drained I'm willing to give in on anything just to get it over with.

You have to realize that different people code differently, we are spending too much time on bikeshedding relatively minor mandane parts of Jenkins, and this is sucking the joy out of my programming. The destructive effect of how this whole thing comes across far outweighs whatever technical gain the actual change may have.

So, thank you very much for your feedback, but no thank you.

stephenc:
@kohsuke There is a design to the credentials API. If you want to integrate with the credentials API then you should integrate in a compatible way with that design. You seem to want to go a different route. You are free to do that if you want, but it is completely contrary to the design and vision of the credentials API that I initially created and have been directing ever since.

The way you want to set this up is going to lead to at best marker interface envy and at worst class envy where users want a credential type that implements multiple credential types. I added support *for migration purposes* to the credentials API to allow for migrating existing credential classes into compatible variants, but it is really designed to allow for graceful migration towards the API design that this pull request leads towards.

Please reconsider, the type envy route that you are advocating is, to my mind, not the way this should go.

> you asked if I wanted you to write it, and this is the only thing I declined

I asked because it was obvious to me that you were not understanding what I was saying, but my understanding is that you said you would write it *because* you understood what I wanted. When I then went and took a look, it was obvious to me that you didn't understand what I was asking, and in the interests of expediency I wrote this pull request to explain the direction that marries with how I designed the credentials API

jglick:
I will reiterate my :+1: on this PR. Aligns well with how we handle SSH-related credentials.

At any rate, I do not think my work is waiting on this either way, since in either `master` or the branch `UsernamePasswordCredentials` is accepted, and should work the same way from the perspective of client code (should be transparent to `materialize`).

recampbell:
What is the status of this PR @reviewbybees?

PR #5, opened 2015-04-13
DockerFingerprints tweaks

jglick:
Making this closer to usable.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
@@ -9,7 +9,7 @@ import hudson.model.Fingerprint;
  * @author Kohsuke Kawaguchi
  */
 public class DockerFromFingerprintFacet extends DockerRunPtrFingerprintFacet {
-    public DockerFromFingerprintFacet(Fingerprint fingerprint, long timestamp) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
```diff
@@ -9,7 +9,7 @@
  * @author Kohsuke Kawaguchi
  */
 public class DockerFromFingerprintFacet extends DockerRunPtrFingerprintFacet {
-    public DockerFromFingerprintFacet(Fingerprint fingerprint, long timestamp) {
-        super(fingerprint, timestamp);
+    DockerFromFingerprintFacet(Fingerprint fingerprint, long timestamp, String imageId) {
```
jglick: Making package-private under the assumption that `DockerFingerprints` covers the legitimate use cases. If not, this could be restored to `public`.

stephenc:
:+1:

tfennelly:
:+1: 

Makes a bit more sense to me now (after looking at Jesse making these changes). But those fingerprint classes could do with some javadoc (using plain lang for guys like me) that explain a bit better what they are for/tracking.

jglick:
> those fingerprint classes could do with some javadoc

Agreed; will leave that to @kohsuke.

tfennelly:
@kohsuke please review this and merge if you're ok with it. Thanks.

PR #6, opened 2015-04-13
[WiP] Simplified DockerClient

tfennelly:
Simplified replacement for https://github.com/jenkinsci/docker-commons-plugin/pull/3

TODO:

- [x] Add test
- [x] Return `build` image ID that can be used for fingerprinting.
- [ ] Change `run` to first do a `docker pull` using `BourneShellScript` (durable)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerClient.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerClient.java
@@ -1,7 +1,7 @@
 /*
  * The MIT License
  *
- * Copyright (c) 2013-2015, CloudBees, Inc.
+ * Copyright (c) 2015, CloudBees, Inc.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
@@ -209,10 +209,10 @@ public class DockerClient {
      */
     public String whoAmI() throws IOException, InterruptedException {
         ByteArrayOutputStream userId = new ByteArrayOutputStream();
-        launcher.launch().cmds("id", "-u").stdout(userId).quiet(true).join();
+        launcher.launch().cmds("id", "-u").stdout(userId).join();
 
         ByteArrayOutputStream groupId = new ByteArrayOutputStream();
-        launcher.launch().cmds("id", "-g").stdout(groupId).quiet(true).join();
+        launcher.launch().cmds("id", "-g").stdout(groupId).join();
 
         return String.format("%s:%s", userId.toString().trim(), groupId.toString().trim());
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/client/DockerClient.java
```diff
@@ -0,0 +1,194 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2015, CloudBees, Inc.
```
KostyaSha: I suppose in 2013 this file didn't exist.

tfennelly: @KostyaSha Funny man.

magnayn:
I don't understand the point of this given that docker-java contains a fully featured java API for invoking docker and building from dockerfiles? 

README.md states the purpose is configuration sharing, not a docker-calling-API ??

src/main/java/org/jenkinsci/plugins/docker/commons/client/LaunchResult.java
```diff
@@ -0,0 +1,67 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2013-2015, CloudBees, Inc.
```
KostyaSha: @tfennelly and here? :laughing:

tfennelly: @KostyaSha The things that keep some people awake at night :wink:

tfennelly:
@magnayn it's not done here yet, but we need to allow [durable execution](https://github.com/jenkinsci/durable-task-plugin) of some of the docker commands for workflow. Atm we're just using a regular `Launcher`. I also note that `docker-java` would drag in a load of dependencies (jersey etc). All that said ... it may end up that we use something like `docker-java` or that none of this code ends up in `docker-common` (and instead, workflow etc use whatever makes sense for them).

KostyaSha:
@tfennelly you can implement workflow-docker and use commons from it with any docker-client implementation you want.

magnayn:
It does bring in dependencies, but that shouldn't particularly be an issue. It has the bigger upside in that, since it communicates directly over http(s), you don't need the docker client utilities to be installed everywhere.

Plus docker-java separates the API from implementation. It'd be perfectly easy to create an implementation that either shelled out or used a non-jaxb web implementation, and allow a build / runtime choice.

jglick:
> It'd be perfectly easy to create an implementation that either shelled out

Possibly so, but I doubt that would work with durable tasks, which generally require a completely different API structure.

tfennelly:
After talking to @jglick we have decided to scrap the idea of putting a Jenkins specific `DockerClient` in here. Closing this PR.

PR #11, opened 2015-04-15
Split KeyMaterial out into KeyMaterialFactory and KeyMaterial

tfennelly:
It is ugly for `KeyMaterial` to be stateful and require a call to `close`.

__NOTE__: The changes here look more than they really are. It's mostly just moving existing code around a bit.

What we're suggesting here is that `KeyMaterial` would still require to be closed, but would only be created and closed at the point where it is needed (and not further back up the call stack). Instead, whatever potentially needs the creds would be given a `KeyMaterialFactory` and if they need the `KeyMaterial` they call `KeyMaterialFactory.materialize` exactly at the point where the creds are needed, and then `KeyMaterial.close` immediately after it finishes using them, making the `KeyMaterial` lifecycle a bit easier to manage.


tfennelly:
@reviewbybees

src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterial.java
```diff
@@ -1,55 +1,69 @@
...
-        if (rhs==null)  return this;
-        return new CompositeKeyMaterial(this,rhs);
-    }
-
+    
     /**
      * {@link KeyMaterial} that does nothing.
      */
-    public static final KeyMaterial NULL = new NullKeyMaterial();
+    public static final KeyMaterial NULL = new KeyMaterial(new EnvVars()) {
```
jglick: Refactor into a `private static` nested class for safe serialization (or `writeReplace` itself to `NULL`).

src/test/java/org/jenkinsci/plugins/docker/commons/SampleDockerBuilder.java
```diff
@@ -39,9 +39,9 @@ public DockerRegistryEndpoint getRegistry() {
     @Override
     public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
         // prepare the credentials to talk to this docker and make it available for docker you'll be forking
-        KeyMaterial key = server.materialize(build);
+        KeyMaterialFactory keyMaterialFactory = server.newKeyMaterialFactory(build).plus(registry.newKeyMaterialFactory(build));
+        KeyMaterial key = keyMaterialFactory.materialize();
         try {
-            key = key.plus(registry.materialize(build));
```
jglick: Good, this is a little simpler. In the former code, though it might not have been apparent, we were handling the case that `server.materialize` succeeded yet `registry.materialize` failed by `close`ing the material from `server.materialize`.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Override
+    public KeyMaterial materialize() throws IOException, InterruptedException {
+        final KeyMaterial lhsMaterialized = lhs.materialize();
+        final KeyMaterial rhsMaterialized;
+        
+        try {
+            rhsMaterialized = rhs.materialize();
+        } catch (Exception e) {
+            lhsMaterialized.close();
```
jglick:
```java
// TODO Java 7+ use chained exceptions
```

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
```diff
@@ -0,0 +1,58 @@
...
+                throw (IOException) e;
+            } else if (e instanceof InterruptedException) {
+                throw (InterruptedException) e;
+            } else {
+                throw new IOException("Error materializing credentials.", e);
+            }
+        }
+
+        EnvVars env = lhsMaterialized.env();
+        env.putAll(rhsMaterialized.env());
```
jglick: What if one of these calls throws an exception?

jglick: I think it suffices to just move the rest of the method inside the `try` block.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
```diff
@@ -0,0 +1,58 @@
...
+
+        return new KeyMaterial(env) {
```
jglick: Not safely serializable, extract to a `private static` nested class.

tfennelly:
@jglick can give this another review too if you get a chance

src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterial.java
```diff
@@ -1,55 +1,73 @@
...
 /**
- * Represents a locally extracted credentials information.
- *
- * <p>
- * Whenever you want to fork off docker directly or indirectly, use this object to set up environment variables
- * so that docker will talk to the right daemon.
- *
- * <p>
- * When you are done using the credentials, call {@link #close()} to allow sensitive information to be removed
- * from the disk.
```
jglick: I think this Javadoc still applies; is there a reason you deleted it?

jglick: I suppose because you moved it all to `KeyMaterialFactory`.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/NullKeyMaterialFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ * @see org.jenkinsci.plugins.docker.commons.KeyMaterial#NULL
+ * @author Kohsuke Kawaguchi
+ */
+@Restricted(NoExternalUse.class)
+public final class NullKeyMaterialFactory extends KeyMaterialFactory {
+
+    // TODO: What's this about and where does it go now? Prob something for Stapler
+    //private Object readResolve() {
+    //    return NULL;
+    //}
```
jglick: Nope, for Java (or XStream) serialization. Should go into `NullKeyMaterial`, so that all instances deserialize back to the singleton `KeyMaterial.NULL`.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/ServerKeyMaterialFactory.java
```diff
@@ -0,0 +1,94 @@
...
+        EnvVars e = new EnvVars();
+        if (host!=null) {
+            e.put("DOCKER_HOST",host);
+        }
+        if (tempCredsDir !=null) {
+            e.put("DOCKER_TLS_VERIFY", "1");
+            e.put("DOCKER_CERT_PATH", tempCredsDir.getRemote());
+        }
+
+        return new KeyMaterial(e) {
```
jglick: Should be a `static` class to be safely serialized.

jglick:
Generally :+1: provided that you can come up with matching PRs for those plugins already using `KeyMaterial` demonstrating that the refactoring (a) works, (b) leaves the client code simpler in some cases (and at least no worse in the others).

src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterial.java
tfennelly: Right, kinda. I should have left those same pointers on `KeyMaterial` though. Will re-add them.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/ServerKeyMaterialFactory.java
tfennelly: Ah bummer ... missed that one. Thanks Jesse.

src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterial.java
tfennelly: Done.

stephenc:
:+1: I need this merged to sort out server key credentials

PR #13, opened 2015-04-16
Turning it into an interface

kohsuke:
IIUC, this change things in the direction Stephen prefers.

@reviewbybees


jglick:
Prefer #4.

kohsuke:
I'm being overridden, and I'm unhappy but it doesn't change the fact that I lost the fight.

PR #14, opened 2015-04-17
Make composition easier

stephenc:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterialContext.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterialContext.java
@@ -1,0 +1,34 @@
+package org.jenkinsci.plugins.docker.commons;
+
+import hudson.FilePath;
+
+import javax.annotation.Nonnull;
+import java.io.Serializable;
+
+/**
+ * Represents the context within a {@link KeyMaterialFactory} can {@link KeyMaterialFactory#materialize()}
+ * {@link KeyMaterial} instances.
+ *
+ * @author Stephen Connolly
+ */
+public class KeyMaterialContext implements Serializable {
+    private static final long serialVersionUID = 1L;
+
+    @Nonnull
+    private final FilePath baseDir;
+
+    public KeyMaterialContext(@Nonnull FilePath baseDir) {
+        this.baseDir = baseDir;
+    }
+
+    /**
+     * Returns the base directory that can be used to {@link KeyMaterialFactory#materialize()}
+     * {@link KeyMaterial} instances.
+     *
+     * @return the base directory.
+     */
+    @Nonnull
+    public FilePath getBaseDir() {
+        return baseDir;
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
@@ -2,10 +2,12 @@ package org.jenkinsci.plugins.docker.commons.impl;
 
 import hudson.EnvVars;
 import org.jenkinsci.plugins.docker.commons.KeyMaterial;
+import org.jenkinsci.plugins.docker.commons.KeyMaterialContext;
 import org.jenkinsci.plugins.docker.commons.KeyMaterialFactory;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
 
+import javax.annotation.Nonnull;
 import java.io.IOException;
 import java.io.Serializable;
 
@@ -22,6 +24,16 @@ public class CompositeKeyMaterialFactory extends KeyMaterialFactory {
         this.factories = factories == null || factories.length == 0
                 ? new KeyMaterialFactory[]{new NullKeyMaterialFactory()}
                 : factories.clone();
+    }
+
+    @Override
+    public synchronized KeyMaterialFactory contextualize(@Nonnull KeyMaterialContext context) {
+        KeyMaterialFactory contextualized = super.contextualize(context);
+        assert contextualized == this;
+        for (KeyMaterialFactory factory : factories) {
+            factory.contextualize(context);
+        }
+        return this;
     }
 
     @Override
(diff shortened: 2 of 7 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
```diff
@@ -10,70 +10,77 @@
...
-                }
-            } finally {
-                if (rhsMaterialized != null) {
-                    rhsMaterialized.close();
+            for (int index = keyMaterials.length - 1; index >= 0; index--) {
+                try {
+                    if (keyMaterials[index] != null) {
+                        keyMaterials[index].close();
+                    }
+                } catch (IOException ioe) {
```
tfennelly: Might want to widen the scope of what gets caught here so as to ensure all are closed.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
```diff
@@ -10,70 +10,77 @@
...
-                lhsMaterialized.close();
-            } finally {
-                rhsMaterialized.close();
+            IOException first = null;
+            for (int index = keyMaterials.length - 1; index >= 0; index--) {
+                try {
+                    if (keyMaterials[index] != null) {
+                        keyMaterials[index].close();
+                    }
+                } catch (IOException ioe) {
```
tfennelly: Same here I think.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
tfennelly: So leaving this catch `IOException` only? Seems like there's a potential leak there.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
```diff
@@ -2,78 +2,109 @@
...
+                try {
+                    if (keyMaterials[index] != null) {
+                        keyMaterials[index].close();
+                    }
+                } catch (IOException e) {
+                    first = first == null ? e : first;
+                } catch (RuntimeException e) {
+                    first = first == null ? e : first;
+                } catch (Throwable e) {
+                    first = first == null ? e : first;
```
tfennelly: Am I blind or are all of those `catch` clauses the same? If so, why not just catch `Throwable`?

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
stephenc: In this case we are safely closing up any materialized instances because of a higher order problem. We want to throw the exception that was the root cause.

tfennelly: Not sure what you mean by that. So if there are multiple `KeyMaterial ` instances and the first one throws a NPE (or some other `RuntimeException`) when calling the `close` here, then the other `KeyMaterial ` instances will not get closed, right? Is that what you want?

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
stephenc: because I like to enumerate the kinds I expect rather than just a blind catch `Throwable`... so that later readers know I am intentionally catching `Throwable` rather than just being lazy

tfennelly: ok :)

tfennelly:
:+1:

src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterialContext.java
```diff
@@ -0,0 +1,34 @@
...
+    /**
+     * Returns the base directory that can be used to {@link KeyMaterialFactory#materialize()}
+     * {@link KeyMaterial} instances.
+     *
+     * @return the base directory.
+     */
+    @Nonnull
+    public FilePath getBaseDir() {
+        return baseDir;
+    }
```
rsandell: Shouldn't ```baseDir``` be removed upon close of the factory?

stephenc: Well that's a question for @tfennelly as the potential leak there exists before this case... we just highlight it better. My thinking is that basedir should be more like a directory that you can create temp stuff in... so it should always exist and thus not need tidy-up. It's the sub-dirs in that base dir that each key material factory is instantiated into that need to tidy-up, which is what `KeyMaterial.close()` does

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
```diff
@@ -2,78 +2,109 @@
...
+            for (int index = keyMaterials.length - 1; index >= 0; index--) {
+                try {
+                    if (keyMaterials[index] != null) {
+                        keyMaterials[index].close();
+                    }
+                } catch (IOException e) {
+                    first = first == null ? e : first;
+                } catch (RuntimeException e) {
+                    first = first == null ? e : first;
+                } catch (Throwable e) {
```
rsandell: All ```Error```s for example ```java.lang.OutOfMemoryError``` are ```Throwable```, are you sure you should be catching them like this?

src/main/java/org/jenkinsci/plugins/docker/commons/KeyMaterialContext.java
rsandell: Ok

tfennelly: @stephenc has it right ... these `KeyMaterial`/`KeyMaterialFactory` classes don't create that dir if I remember, so they should not be cleaning it up. They should only clan up dirs that they were responsible for creating. If something else created the `baseDir` then it is responsible for cleaning it up.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/CompositeKeyMaterialFactory.java
stephenc: So if there is an error we should not try and clean up any secrets left on the disk then?

rsandell: We might not be able to, as we are out of memory

rsandell:
:+1:

PR #15, opened 2015-04-17
Server credentials

stephenc:


stephenc:
@reviewbybees in case you want to follow along

stephenc:
@reviewbybees

I need to refactor this again once we get a concrete credentials type upstream that is appropriate for this use case.

The PKCS12 certificate credential type (which is what `CertificateCredentialsImpl` is, as it stores the certificate(s) and key in a PKCS12 `KeyStore`) does not seem to like letting us retrieve the key entry which significantly complicates life for the user.

I suspect the way to handle this issue is to use a password to store the keys in the PKCS12 `KeyStore`, but without testing that may complicate life for the user.

In principle we should be able to just add an entry into the key store with the certificate chain, and thus retrieve the entry and it's chain, but as the docker client does not seem to currently support passwords on the client key... well complexity

PR #16, opened 2015-04-21
Add equals() and hashcode() methods to endpoint classes

oleg-nenashev:
Methods are required to use EndPoints as a key in hash maps

@stephenc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
@@ -159,6 +159,32 @@ public class DockerRegistryEndpoint extends AbstractDescribableImpl<DockerRegist
         return "DockerRegistryEndpoint[" + url + ";credentialsId=" + credentialsId + "]";
     }
 
+    @Override
+    public int hashCode() {
+        int hash = 3;
+        hash = 31 * hash + (this.url != null ? this.url.hashCode() : 0);
+        hash = 31 * hash + (this.credentialsId != null ? this.credentialsId.hashCode() : 0);
+        return hash;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final DockerRegistryEndpoint other = (DockerRegistryEndpoint) obj;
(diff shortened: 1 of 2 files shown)


rsandell:
:+1:

src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
```diff
@@ -159,6 +159,32 @@ public String imageName(String userAndRepo) throws IOException {
...
+    public boolean equals(Object obj) {
+        if (obj == null) {
+            return false;
+        }
+        if (getClass() != obj.getClass()) {
+            return false;
+        }
+        final DockerRegistryEndpoint other = (DockerRegistryEndpoint) obj;
+        if ((this.url == null) ? (other.url != null) : !this.url.equals(other.url)) {
+            return false;
```
jglick: Can be simplified using `ObjectUtils.equals` (Commons Lang) or `Objects.equal` (Guava).

jglick: e.g.

```java
@Override public boolean equals(Object obj) {
    return obj instanceof DockerRegistryEndpoint &&
           ObjectUtils.equals(url, ((DockerRegistryEndpoint) obj).url) &&
           ObjectUtils.equals(credentialsId, ((DockerRegistryEndpoint) obj).credentialsId);
}
```

jglick:
:+1:

src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
oleg-nenashev: @jglick 
Sorry, I've missed the comment. <code>ObjectUtils</code> is deprecated by <code>java.util.Objects</code>, but it does not change the comment's idea to reuse the existing code. This idea is definitely good. Unfortunately, all methods in classes fall-back to <code>A.equals(B)</code>.

There could be a generic reflection-based Lib method, but a manual implementation is faster, which is significant for the class usage in containers. I would propose to leave the change as is.

jglick: > `ObjectUtils` is deprecated by `java.util.Objects`

Yes, once we are on Java 7+.

This was not important, just a thought.

PR #17, opened 2015-04-27
Added imageId on ContainerRecord

tfennelly:


tfennelly:
@reviewbybees

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/ContainerRecord.java
```diff
@@ -20,14 +21,16 @@
     
     private final String host;
     private final String containerId;
+    private final String imageId;
```
jglick: `transient`

jglick: `DockerRunPtrFingerprintFacet` already has this. `DockerRunFingerprintFacet` should have a `readResolve` that sets it back on the records.

tfennelly:
@jglick test is simplified now. Not using `TestBuilder` but is testing the `readResolve` part enough I think. Not a great test, but better than nothing.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/ContainerRecord.java
```diff
@@ -48,6 +51,22 @@ public String getContainerId() {
...
+     */
+    public String getImageId() {
+        return imageId;
+    }
+
+    /**
+     * Set the image ID of the image from which this container was started.
+     * @param imageId The image ID of the image from which this container was started.
+     */
+    public void setImageId(@Nonnull String imageId) {
```
oleg-nenashev: I doubt a public setter method is required

jglick: It would be if you wanted to make the equivalent of `DockerRunFingerprintFacet` somewhere else.

PR #18, opened 2015-04-28
DockerServerCredentials is now concrete, so we can test its use

jglick:
Should have been part of #15 but was forgotten I guess. @reviewbybees


oleg-nenashev:
:+1:

PR #19, opened 2015-04-29
Cleaning up DockerFingerprints.addRunFacet method sig

tfennelly:
Removing some duplication.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
@@ -27,18 +27,22 @@ public class DockerFingerprints {
     /**
      * Gets {@link Fingerprint} for a given docker image.
      */
-    public static @CheckForNull Fingerprint of(String imageId) throws IOException {
+    public static @CheckForNull Fingerprint of(@Nonnull String imageId) throws IOException {
         return Jenkins.getInstance().getFingerprintMap().get(trim(imageId));
     }
 
-    private static @Nonnull Fingerprint make(Run<?,?> run, String imageId) throws IOException {
+    private static @Nonnull Fingerprint make(@Nonnull Run<?,?> run, @Nonnull String imageId) throws IOException {
         return Jenkins.getInstance().getFingerprintMap().getOrCreate(run, "<docker-image>", trim(imageId));
     }
 
     /**
      * Adds a new {@link ContainerRecord} for the specified image, creating necessary intermediate objects as it goes.
      */
-    public static void addRunFacet(String imageId, ContainerRecord record, Run<?,?> run) throws IOException {
+    public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?> run) throws IOException {
+        String imageId = record.getImageId();
+        if (imageId == null) {
+            throw new IllegalArgumentException("Invalid ContainerRecord instance. ContainerRecords passed to addRunFacet must have an image ID.");
+        }
         Fingerprint f = make(run, imageId);
         Collection<FingerprintFacet> facets = f.getFacets();
         DockerRunFingerprintFacet runFacet = null;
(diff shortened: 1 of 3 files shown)


src/test/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerRunFingerprintFacetTest.java
```diff
@@ -46,14 +46,17 @@ public void test_readResolve() throws Exception {
...
+        DockerFingerprints.addRunFacet(r1, b);
 
-        Assert.assertNull(r.getImageId());
+        Fingerprint fingerprint = DockerFingerprints.of(IMAGE_ID);
+        DockerRunFingerprintFacet facet = new DockerRunFingerprintFacet(fingerprint, System.currentTimeMillis(), IMAGE_ID);
+        ContainerRecord r2 = new ContainerRecord("192.168.1.10", "cid", null, "magic", System.currentTimeMillis(), Collections.<String, String>emptyMap());
+        facet.add(r2);
+        
+        Assert.assertNull(r2.getImageId());
         facet.readResolve();
```
jglick: Would still prefer a more natural test here, that after saving the record and running `rule.jenkins.reload()` you get everything back, but am not going to insist on it.

jglick:
:+1:

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -27,18 +27,22 @@ private static String trim(String imageId) {
...
         return Jenkins.getInstance().getFingerprintMap().getOrCreate(run, "<docker-image>", trim(imageId));
     }
 
     /**
      * Adds a new {@link ContainerRecord} for the specified image, creating necessary intermediate objects as it goes.
      */
-    public static void addRunFacet(String imageId, ContainerRecord record, Run<?,?> run) throws IOException {
+    public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?> run) throws IOException {
+        String imageId = record.getImageId();
+        if (imageId == null) {
```
oleg-nenashev: Javadoc and annotations are missing for <code> ContainerRecord::getImageId()</code>. Is there a real need to have such check here? It worths just to make the field <code>@Nonnull</code>

jglick: True, I think it would suffice for the contract to be defined in `ContainerRecord` itself, and just assumed non-null here.

src/test/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerRunFingerprintFacetTest.java
tfennelly: That would be better, but I don't know what I need to do to mimic that. I'm sure I'd figure it out, but I suppose there's a bit of a tradeoff in terms of the amount of time I have to spend writing a test Vs the level of potential risk.

Personally, I wouldn't have opened the `transient` + `readResolve` can of worms at all.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
tfennelly: I suppose it depends on how the XStream deserialization works, and that was why I did not add the `@Nonnull` annotation on the constructor for `imageId` (remember that `ContainerRecord.imageId` is `transient`).

@oleg-nenashev `ContainerRecord.getImageId` and `ContainerRecord.setImageId` both have Javadoc. Or did you mean something else?

oleg-nenashev: @tfennelly 
The plugin has not been officially released yet,  so I'm not sure what's the problem with serialization. Every field may be null after the serialization, but replacing one runtime exception by another one won't help. imageId does not differ from containerId in such scope.

Off-topic: XStream could check annotations and fail on violations

jglick: `imageId` is expected to be non-null at all times except _immediately and temporarily_ after deserialization, so I think `@Nonnull` is appropriate.

jglick:
:+1:

kohsuke:
:+1:

PR #20, opened 2015-04-30
Docker Fingerprint Actions

oleg-nenashev:
This action allows to...
* Navigate to Docker fingerprints from Runs
* Quickly search fingerprints by Run

screen shot 2015-04-30 at 04 26 52

Plan:
- [X] - Working prototype
- [x] Better GUI
- [x] Unit tests

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
@@ -1,0 +1,132 @@
+/*
+ * The MIT License
+ *
+ * Copyright (c) 2015, CloudBees, Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.jenkinsci.plugins.docker.commons.fingerprint;
+
+import hudson.model.Fingerprint;
+import hudson.model.Run;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import javax.annotation.CheckForNull;
+import jenkins.model.RunAction2;
+import org.jenkinsci.plugins.docker.commons.Messages;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+/**
+ * Adds a link, which provides a list of fingerprints with
+ * {@link DockerRunPtrFingerprintFacet}s.
+ *
+ * @author Oleg Nenashev
+ */
+public class DockerFingerprintAction implements RunAction2 {
+
+    private final Set<String> imageIDs;
+    transient Run<?, ?> run;
+
+    public DockerFingerprintAction() {
+        this.imageIDs = new HashSet<String>();
+    }
+
+    public Run<?, ?> getRun() {
+        return run;
+    }
+
+    @Override
+    public void onAttached(Run<?, ?> r) {
+        run = r;
+    }
+
+    @Override
+    public void onLoad(Run<?, ?> r) {
+        run = r;
+    }
+
+    @Override
+    public String getIconFileName() {
+        return "fingerprint.png";
+    }
+
+    @Override
+    public String getDisplayName() {
+        return Messages.DockerFingerprintAction_DisplayName();
+    }
+
+    @Override
+    public String getUrlName() {
+        return "docker-fingerprints";
+    }
+
+    /**
+     * Gets a list of fingerprints related to the action
+     *
+     * @return Unmodifiable set of items.
+     */
+    public Set<String> getImageIDs() {
+        return Collections.unmodifiableSet(imageIDs);
+    }
+
+    @Restricted(NoExternalUse.class)
+    public @CheckForNull String getFingerprintHash(@CheckForNull String imageId) {
+        return (imageId != null) ? DockerFingerprints.getImageFingerprintHash(imageId) : null;
+    }
+    
+    @Restricted(NoExternalUse.class)
+    public @CheckForNull Fingerprint getFingerprint(@CheckForNull String imageId) {
+        if (imageId == null) {
+            return null;
+        }
+        
+        try {
+            return DockerFingerprints.of(imageId);
+        } catch (IOException ex) {
+            return null; // nothing to do in web UI - return null as well
+        }
+    }
+
+    /**
+     * Adds an action with a reference to fingerprint if required. It's
+     * recommended to call the method from {
+     *
+     * @BulkChange} transaction to avoid saving the {@link Run} multiple times.
+     * @param fp Fingerprint
+     * @param imageId ID of the docker image
+     * @param run Run to be updated
+     * @throws IOException Cannot save the action
+     */
+    static void addToRun(Fingerprint fp, String imageId, Run run) throws IOException {
+        synchronized (run) {
+            DockerFingerprintAction action = run.getAction(DockerFingerprintAction.class);
+            if (action == null) {
+                action = new DockerFingerprintAction();
+                run.addAction(action);
+            }
+            
+            if (action.imageIDs.add(imageId)) {
+                run.save();
+            } // else no need to save updates
+        }
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
@@ -12,7 +12,7 @@ import javax.annotation.Nonnull;
 import jenkins.model.FingerprintFacet;
 
 /**
- * Entry point into fignerprint related functionalities in Docker.
+ * Entry point into fingerprint related functionalities in Docker.
  */
 public class DockerFingerprints {
     private DockerFingerprints() {} // no instantiation
@@ -22,6 +22,15 @@ public class DockerFingerprints {
             throw new IllegalArgumentException("Expecting 64char full image ID, but got " + imageId);
         }
         return imageId.substring(0, 32);
+    }
+    
+    /**
+     * Gets a fingerprint hash for Docker image.
+     * @param imageId Docker image ID.
+     * @return 32-symbol fingerprint hash
+     */
+    /*package*/ static @Nonnull String getImageFingerprintHash(@Nonnull String imageId) {
+        return trim(imageId);
     }
 
     /**
@@ -57,6 +66,7 @@ public class DockerFingerprints {
             }
             runFacet.add(record);
             runFacet.addFor(run);
+            DockerFingerprintAction.addToRun(f, imageId, run);
             bc.commit();
         } finally {
             bc.abort();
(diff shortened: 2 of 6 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,94 @@
+package org.jenkinsci.plugins.docker.commons.fingerprint;
```
jglick: License header.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,94 @@
...
+ * Adds a link, which provides a list of fingerprints with {@link DockerRunPtrFingerprintFacet}s.
+ * @author Oleg Nenashev
+ */
+public class DockerFingerprintAction implements RunAction2 {
+
+    private final Set<String> fingerprintKeys;
+    transient @CheckForNull Run<?, ?> run;
+
+    public DockerFingerprintAction(Run<?, ?> run) {
+        this.run = run;
```
jglick: For a `RunAction2` there is no need to pass the build in the constructor. `onAttached` will be called during `Run.addAction`.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
oleg-nenashev: It is missing everywhere, so I've just foo]llowed the style

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,94 @@
...
+                run.addAction(action);
+            }
+
+            if (!action.getFingerprintKeys().contains(imageId)) {
+                action.getFingerprintKeys().add(imageId);
+            } else {
+                return; // no need to save updates
+            }
+
+            run.save();
```
jglick: If you like, this can be shortened a bit:

```java
if (action.getFingerprintKeys().add(imageId)) {
    run.save();
} // else no need to save updates
```

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,94 @@
...
+        return Messages.DockerFingerprintAction_DisplayName();
+    }
+
+    @Override
+    public String getUrlName() {
+        return "docker-fingerprints";
+    }
+
+    public Set<String> getFingerprintKeys() {
+        return fingerprintKeys;
```
jglick: Wise to have the `public` method return `Collections.immutableSet` (if not a copy-on-write implementation), and access the field directly from `addToRun`. I suppose that this method is considered an API that other plugins might need.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,94 @@
...
+     * @param imageId ID of the docker image
+     * @param run Run to be updated
+     * @throws IOException Cannot save the action
+     */
+    static void addToRun(Fingerprint fp, String imageId, Run run) throws IOException {
+        synchronized (run) {
+            DockerFingerprintAction action = run.getAction(DockerFingerprintAction.class);
+            if (action == null) {
+                action = new DockerFingerprintAction(run);
+                run.addAction(action);
```
jglick: BTW you can use `BulkChange` to avoid saving the `Run` twice.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,94 @@
...
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+
+/**
+ * Adds a link, which provides a list of fingerprints with {@link DockerRunPtrFingerprintFacet}s.
+ * @author Oleg Nenashev
+ */
+public class DockerFingerprintAction implements RunAction2 {
+
+    private final Set<String> fingerprintKeys;
+    transient @CheckForNull Run<?, ?> run;
```
jglick: It is always nonnull except for very temporarily, and in practice no one is going to check for it being null, so this annotation is misleading.

src/main/resources/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction/index.jelly
```diff
@@ -0,0 +1,75 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+<j:jelly xmlns:j="jelly:core" xmlns:st="jelly:stapler" xmlns:d="jelly:define" 
+         xmlns:l="/lib/layout" xmlns:t="/lib/hudson" xmlns:f="/lib/form" xmlns:i="jelly:fmt">
+  <l:layout title="${%title(it.run.fullDisplayName)}">
```
jglick: `permission="${it.run.ARTIFACTS}"`

jglick: (or perhaps `it.run.parent.READ`, depending on what you think the threat is)

jglick:
We will presumably want to distinguish `FROM` fingerprints from others, though this is not fleshed out at the code level anyway.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
oleg-nenashev: agreed

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
oleg-nenashev: BulkChange is being called in upper calls, but it definitely makes sense to recommend such implementation in Javadoc at least

src/main/resources/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction/index.jelly
oleg-nenashev: Seems we'll need a new permission group for Docker at some point. So we'll be able to create a new permission inheriting from <code>READ</code> there

jglick: > we'll need a new permission group for Docker at some point

We will? Why?

(We already have lots of specialized permissions and permission groups floating around in various plugins so I think there should be a high bar to adding more, lest administrators configuring the authorization strategy be totally confused.)

oleg-nenashev:
@jglick 
+1 for the categorization. The simplest way would be to add <code>getDisplayName()</code> defaulting to the class name, but we can also implement several content contributors (extension points, etc.). I would start from the display name.

src/main/resources/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction/index.jelly
oleg-nenashev: > (We already have lots of specialized permissions and permission groups floating around in various plugins so I think there should be a high bar to adding more, lest administrators configuring the authorization strategy be totally confused.)

+1. Unfortunately, there's no other mechanism, which could be easily extended for security check purposes. Fingerprint index page does not enforce additional permissions, so I'd follow the same approach with actions.

oleg-nenashev:
@jglick 
> We will presumably want to distinguish FROM fingerprints from others, though this is not fleshed out at the code level anyway.

I suppose that Docker image fingerprint will finally store every available info about the image. We could    publish a list of <code>Facet</code>s in the additional column, but seems we cannot do more at the current stage.

Does it worth to merge the current PR to the alpha version of docker-commons?

jglick:
#21 actually did flesh out `DockerFromFingerprintFacet`, so if you have not seen that already you should consider integrating with it.

oleg-nenashev:
@jglick 
The last commit has been rebased to the new version with #21

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,141 @@
...
+    private Set<String> imageIDs;
+    transient Run<?, ?> run;
+
+    public DockerFingerprintAction() {
+        this.imageIDs = new HashSet<String>();
+    }
+    
+    public Object readResolve() {
+        if (imageIDs == null) {
+            imageIDs = new HashSet<String>();
```
jglick: Is this actually needed? You would add such a `readResolve` method (typically `private`, maybe `protected`) when you introduce a new field for which a null/0 default value does not make sense.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -0,0 +1,141 @@
...
+        synchronized (run) {
+            DockerFingerprintAction action = run.getAction(DockerFingerprintAction.class);
+            if (action == null) {
+                action = new DockerFingerprintAction();
+                run.addAction(action);
+            }
+            
+            if (action.imageIDs.add(imageId)) {
+                run.save();
+            } // else no need to save updates
```
jglick: Can use `BulkChange` to save only once, but probably not important.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -23,6 +23,15 @@ private static String trim(String imageId) {
...
         return imageId.substring(0, 32);
     }
+    
+    /**
+     * Gets a fingerprint hash for Docker image.
+     * @param imageId Docker image ID.
+     * @return 32-symbol fingerprint hash
+     */
+    /*package*/ static @Nonnull String getImageFingerprintHash(@Nonnull String imageId) {
+        return trim(imageId);
```
jglick: Why not just make `trim` package-private.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -57,6 +66,7 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
             }
             runFacet.add(record);
             runFacet.addFor(run);
+            DockerFingerprintAction.addToRun(f, imageId, run);
```
jglick: Would be simpler to move this call inside `addFor`.

jglick:
Minor comments but :+1: anyway.

My mention of #21 was in the sense that it might make sense for the UI to display different kinds of facets differently, but I think that can wait.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
oleg-nenashev: <code>trim()</code> has a weird name for public methods. BTW seems it makes sense to have a public method for external users, who may want to convert imageId to hash w/o retrieving <code>Fingerprint</code>s. It's a useful approach due to performance reasons.

jglick: I do not care much about the name; anyway the proposed new method is not `public` either. Just commenting that it is odd to have _both_ methods.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
oleg-nenashev: In the current plugin implementation <code>BulkChange</code> is being created on the upper level

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
oleg-nenashev: I'll rework it in a separate pull request, because it needs an additional discussion regarding the public access modifier.

oleg-nenashev:
Fixed minor comments excepting the ones discussed.
@reviewbybees, any additional comments?

jglick:
Argh, please do not `rebase` within a PR. Now I cannot see which “minor fixes” you made, and have to reread the whole thing.

(Someone _merging_ a PR can rebase if they really feel the need to. I would not.)

jglick:
:+1:

oleg-nenashev:
@jglick 
I've just got rid of <code>readResolve()</code>. I'll try to avoid rebases in the future

PR #21, opened 2015-04-30
Make DockerFromFingerprintFacet track the linkage between base and derived images

jglick:
- [X] write API
- [x] prove it works as intended
- [x] refactor to two `Facet`s
- [x] verify/fix amalgamation

@reviewbybees


src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
```diff
@@ -3,13 +3,30 @@
...
+ * As a facet on a fingerprint corresponding to the base image, refers to the derived image.
+ * As a facet on a fingerprint corresponding to the derived image, refers to the base image.
+ * Both fingerprints should be associated with the same build.
+ * @see DockerFingerprints#addFromFacet
  */
 public class DockerFromFingerprintFacet extends DockerRunPtrFingerprintFacet {
-    DockerFromFingerprintFacet(Fingerprint fingerprint, long timestamp, String imageId) {
+
+    private final String otherImageId;
+    private final boolean inverse;
```
oleg-nenashev: Probably it's an overkill

jglick: Well otherwise you have no way of (efficiently) going in both directions. Since I am not totally sure what the use cases are for this facet it seemed prudent to retain all the relevant information: what was built `FROM` a given base image, and what base image a given image was built `FROM`.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
```diff
@@ -3,13 +3,30 @@
...
 
 /**
- * As a facet on a fingerprint that represents a docker image,
- * indicates that another image was produced by deriving from it. See "FROM" in Dockerfile.
- *  
- * @author Kohsuke Kawaguchi
+ * Facet representing the fact that one Docker image was derived from another.
+ * This corresponds to the {@code FROM} directive in a {@code Dockerfile}.
+ * As a facet on a fingerprint corresponding to the base image, refers to the derived image.
+ * As a facet on a fingerprint corresponding to the derived image, refers to the base image.
```
tfennelly: So this is to support base <-> derived bi-directional search? Could it be worded more clearly for dummies like me?

tfennelly:
:+1: 

Funky :smile:

tfennelly:
I'm not sure how fingerprinting (and search nav of) works, but if it was possible to search "downward" from each build then can't you get the full picture in both directions? Just wondering.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -66,7 +66,12 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
     /**
      * Creates a new {@link DockerFromFingerprintFacet} and adds a run. Or adds to an existing one.
      */
-    public static void addFromFacet(@Nonnull String imageId, @Nonnull Run<?,?> run) throws IOException {
+    public static void addFromFacet(@Nonnull String baseImageId, @Nonnull String derivedImageId, @Nonnull Run<?,?> run) throws IOException {
```
jtnord: This certainly deserves javadoc on the parameters.
baseImageId can be `null` if you create a [base](https://docs.docker.com/terms/image/#base-image) [image](https://docs.docker.com/articles/baseimages/)

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -66,7 +66,12 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
...
+        long timestamp = System.currentTimeMillis();
+        addHalfFromFacet(baseImageId, derivedImageId, false, run, timestamp);
+        addHalfFromFacet(derivedImageId, baseImageId, true, run, timestamp);
+    }
+    private static void addHalfFromFacet(@Nonnull String imageId, @Nonnull String otherImageId, boolean inverse, @Nonnull Run<?,?> run, long timestamp) throws IOException {
```
jtnord: needs some comments on what this is doing and why because at first glance it's not obvious.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
```diff
@@ -3,13 +3,30 @@
 import hudson.model.Fingerprint;
 
 /**
- * As a facet on a fingerprint that represents a docker image,
- * indicates that another image was produced by deriving from it. See "FROM" in Dockerfile.
- *  
- * @author Kohsuke Kawaguchi
+ * Facet representing the fact that one Docker image was derived from another.
+ * This corresponds to the {@code FROM} directive in a {@code Dockerfile}.
+ * As a facet on a fingerprint corresponding to the base image, refers to the derived image.
```
jtnord: terminology here is wrong.  A base image is the root of all the parents parent.  I think you mean the parent image?

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
jtnord: these two sentences above just mess with my head and I can't understand what you are trying to tell people.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
```diff
@@ -3,13 +3,30 @@
...
+    private final String otherImageId;
+    private final boolean inverse;
+
+    DockerFromFingerprintFacet(Fingerprint fingerprint, long timestamp, String imageId, String otherImageId, boolean inverse) {
         super(fingerprint, timestamp, imageId);
+        this.otherImageId = otherImageId;
+        this.inverse = inverse;
+    }
+
+    public String getBaseImageId() {
```
jtnord: its not baseImage.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
```diff
@@ -3,13 +3,30 @@
...
+        return inverse ? otherImageId : getImageId();
```
jtnord: this messes with my head - when I am asking a facet for a parent I should always get its parent not itself surely?
if it doesn't know then it should return null or throw an exception?

jtnord:
There are certainly terminology inaccuracies that need addressing - but I also have some open issues on the code  :-1:

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: Agreed.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
```diff
@@ -3,13 +3,30 @@
...
     }
+
+    public String getDerivedImageId() {
+        return inverse ? getImageId() : otherImageId;
```
jtnord: How to these get amalgamater.  One parent can have many children.  (is this just that a single fingerprint has many Facets?)

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: Agreed that using the term “base image” is misleading in this context. The problem is not helped by the fact that Docker documentation [also uses this term in relation to the `FROM` instruction](https://docs.docker.com/reference/builder/#from).

“Parent image” is not exactly right for the first parameter here, though. It could in fact be the great-great-grandparent:

```
FROM java:openjdk-7u65-jdk
COPY something /tmp
COPY something-else /tmp
COPY yet-again /tmp
COPY still-more /tmp
```

Here the second parameter should represent the image with `/tmp/still-more`. The parent lacks this but has `/tmp/yet-again`. What we want for the first parameter is the ID of `java:openjdk-7u65-jdk` (not a base image in this case, but could be if we chose for example `ubuntu`—does not matter).

So perhaps it should just be called `fromImageId`. I suppose it could be permitted to be null, in the rare case you were actually creating a base image from a Jenkins job, using

```
FROM scratch
```

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
jglick: Was trying to be explicit to avoid confusion but I guess I failed!

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
jglick: Do not understand the question you are asking. `getImageId()` is always the image whose fingerprint this facet is attached to. Depending on the mode of the facet, that could be the ancestor image, or the derived image.

Perhaps it would be less confusing to remove the mode flag and create two `Facet` classes, with identical code other than field/method names?

jglick:
> if it was possible to search "downward" from each build

Not sure what you are suggesting.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
jtnord: +1 to separate classes - but I think they should have a different method signature depending on if its pointing at the from/parent or the derived/child

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFromFingerprintFacet.java
jglick: Good question, I need to test that case.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerAncestorFingerprintFacet.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.annotation.Nonnull;
+
+/**
+ * Facet representing the fact that some Docker image was derived from this one.
+ * This facet would be added to the fingerprint corresponding to the {@code FROM} instruction in a {@code Dockerfile}.
+ * {@link #getDescendantImageId} indicates the image that was built from that file.
+ * @see DockerFingerprints#addFromFacet
+ * @see DockerDescendantFingerprintFacet
+ */
+public class DockerAncestorFingerprintFacet extends DockerRunPtrFingerprintFacet {
```
jtnord: I don't get why this is the Ancestor Facet that points to a descendant
(and the Descendant facet points to the Ancestor).

i.e. Why doesn't the descendant Facet point to the descendant image and the Ancestor facet point to the Ancestor?

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -64,25 +64,54 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
     }
 
     /**
-     * Creates a new {@link DockerFromFingerprintFacet} and adds a run. Or adds to an existing one.
+     * Creates a new {@link DockerAncestorFingerprintFacet} and {@link DockerDescendantFingerprintFacet} and adds a run.
+     * Or adds to existing facets.
+     * @param ancestorImageId the ID of the image specified in a {@code FROM} instruction, or null in case of {@code scratch} (i.e., the descendant is a base image)
+     * @param descendantImageId the ID of the image which was built
```
jtnord: nit: isn't this just imageId

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerAncestorFingerprintFacet.java
```diff
@@ -0,0 +1,32 @@
...
+
+    final Set<String> descendantImageIds = new TreeSet<String>();
```
jtnord: would prefer that access to this was hidden and a helper addDescenantImage(String id) was added.

Also TreeSet is not synchronized and can be called from multiple threads IIUC

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerAncestorFingerprintFacet.java
jglick: Well, the ancestor facet is attached to the fingerprint of the ancestor.

jtnord: Whilst it may be in this code - if you actually have an arbitrary `Fingerprint` that fingerprint may have both `DockerAncestorFingerprintFacet` and a `DockerDescendantFingerprintFacet`.  At this point I guess the naming is a littl confused as the object of the noun isn't there.  I guess its a `DockerDecendantOfFingerPrintFacet` and a `DockerAncenstorOfFingerprintFacet` - rather than my thinking that the Fingerprint has an ancestor facet that would point to the ancestor.

I'm willing to be swayed if others also shoot me down :smile:

jglick: I tried to make it as clear as possible what it means in the summary sentence. Not sure what else you are suggesting.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: Do not understand the comment.

jtnord: `descendantImageId` the "descendant" part is redundant and potentially confusing.
if you have a descendant and an ancestor then the questions that comes to mind is who is in the descendant's parent and who is the ancestors child; (that is - it implies something is in the middle  A -> B -> C where A is the ancestor, B is the thing who you are inferring relationships about and C is the descendant).

Like I said - NIT.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerAncestorFingerprintFacet.java
jtnord: I think I am just suggesting adding an "of" to the name to show its an AncestorOf something rather than the AncestorFacet pointing to the ancestor Of this image.

YMMV - others may disagree - I am willing to be convinced otherwise.

jglick: `DockerAncestorOfFingerprintFacet` does not read well. It sounds like something is an ancestor of the fingerprint!

I have reread this several times but still find the original names sounding right to me.

jtnord:
I am happy with the code - and if others don't feel the same about the naming of the classes and parameters then I am happy to be overruled on this

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -64,25 +64,54 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
...
      */
-    public static void addFromFacet(@Nonnull String imageId, @Nonnull Run<?,?> run) throws IOException {
-        Fingerprint f = make(run, imageId);
+    public static void addFromFacet(@CheckForNull String ancestorImageId, @Nonnull String descendantImageId, @Nonnull Run<?,?> run) throws IOException {
+        long timestamp = System.currentTimeMillis();
+        if (ancestorImageId != null) {
+            Fingerprint f = make(run, ancestorImageId);
+            Collection<FingerprintFacet> facets = f.getFacets();
+            DockerAncestorFingerprintFacet ancestorFacet = null;
+            for (FingerprintFacet facet : facets) {
```
oleg-nenashev: Makes sense to finally decouple this code snippet somewhere.
E.g. 
```
public static @CheckForNull @SuppressWarnings("unchecked")
            <TFacet extends FingerprintFacet> TFacet getFacet
            (@Nonnull Fingerprint fingerprint, Class<TFacet> facetClass) {  
        for ( FingerprintFacet facet : fingerprint.getFacets()) {
            if (facetClass.isAssignableFrom(facet.getClass())) {
                return (TFacet)facet;
            }
        }
        return null;      
    }
```

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -64,25 +64,54 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
...
         try {
-            if (fromFacet == null) {
-                fromFacet = new DockerFromFingerprintFacet(f, System.currentTimeMillis(), imageId);
-                facets.add(fromFacet);
+            if (descendantFacet == null) {
+                descendantFacet = new DockerDescendantFingerprintFacet(f, timestamp, descendantImageId);
+                facets.add(descendantFacet);
             }
-            fromFacet.addFor(run);
+            descendantFacet.addAncestorImageId(ancestorImageId);
```
oleg-nenashev: <code>ancestorImageId</code>  may be <code>null</code> here. Does it make sense to add nulls to the facet?

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: > Makes sense to finally decouple this code snippet somewhere.

Yes a `Fingerprint.getFacet` method would be appreciated.

kohsuke:
I found myself resonating with @jtnord on the name. That is, `DockerAncestorFingerprintFacet` should be the facet on the descendant that points to ancestors.

Jesse was kind enough to be overridden on that one, and so here is my :+1: , thanks Jesse!

kohsuke:
:+1:

jtnord:
:+1:

oleg-nenashev:
:+1:

PR #22, opened 2015-05-01
Javadoc refinement relating to base images

jglick:
https://github.com/jglick/docker-commons-plugin/commit/91fdceac63148f267141bf681a2b8726770df6d0 in #21 led to the situation that the ancestor ID set is actually empty.

It would have been nice if Docker assigned the ID `0000000000000000000000000000000000000000000000000000000000000000` (or anything, really) to the empty image. But it does not seem to have a concept of zero:

```
$ docker inspect -f {{.Parent}} ubuntu

$ docker inspect scratch
Error: No such image or container: scratch
[]
```

The choices here are:

* Use `null` as the parameter to `addFromFacet`, as I have done, but do not add it to `DockerAncestorFingerprintFacet.ancestorImageIds`.
* Do not even create the `DockerAncestorFingerprintFacet` (but if it already exists, leave its `ancestorImageIds` unmodified).
* Add a `null` to the set, and make callers check for null.
* Use the empty string, and make callers check `.isEmpty()` before assuming an “ID” is really a 64-digit hash.
* Adopt the convention of a zero-padded hash for `scratch`, knowing that `docker-inspect` will not produce it.
* Not support fingerprinting of base images at all.

None seem attractive, but @reviewbybees


oleg-nenashev:
:+1:

PR #23, opened 2015-05-05
Prevent IllegalStateException when reusing the NullKeyMaterialFactory

carlossg:
If two calls are made to contextualize from two different places with empty credentials, ie. DockerServerEndpoint and DockerRegistryEndpoint, they both reuse the same object and the second call gets IllegalStateException

@reviewbybees


oleg-nenashev:
:+1:

jglick:
@stephenc it seems your API in ccfd7fe8896873fff599ab02ee0bdc27faaab6cf is causing problems.

The problem here is arguably not a missing override of `contextualize`, which was not really intended to be overridable, but the fact that `NULL` is a singleton when a new `NullKeyMaterialFactory` should rather be created each time. Alternately, that `contextualize` is a poor API; there should be two types, an actual factory, and a factory-with-context.

At any rate, :+1: for this fix in isolation.

PR #24, opened 2015-05-05
Add a way to parse repo/namespace/image into a DockerRegistryEndpoint

carlossg:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
@@ -6,6 +6,7 @@ import com.cloudbees.plugins.credentials.common.StandardCredentials;
 import com.cloudbees.plugins.credentials.common.StandardListBoxModel;
 import com.cloudbees.plugins.credentials.domains.DomainRequirement;
 import com.cloudbees.plugins.credentials.domains.HostnameRequirement;
+
 import hudson.Extension;
 import hudson.Util;
 import hudson.model.AbstractBuild;
@@ -16,18 +17,22 @@ import hudson.model.Item;
 import hudson.remoting.VirtualChannel;
 import hudson.util.ListBoxModel;
 import jenkins.authentication.tokens.api.AuthenticationTokens;
-import hudson.util.ListBoxModel;
 import jenkins.model.Jenkins;
+
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
+
 import java.io.IOException;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.util.Collections;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import static com.cloudbees.plugins.credentials.CredentialsMatchers.*;
 
@@ -41,6 +46,18 @@ import static com.cloudbees.plugins.credentials.CredentialsMatchers.*;
  * @author Kohsuke Kawaguchi
  */
 public class DockerRegistryEndpoint extends AbstractDescribableImpl<DockerRegistryEndpoint> {
+
+    /**
+     * Some regex magic to parse docker registry:port/namespace/name:tag into parts, using the same constraints as
+     * docker push.
+     * 
+     * registry must not contain / and must contain a .
+     * 
+     * everything but name is optional
+     */
+    private static final Pattern DOCKER_REGISTRY_PATTERN = Pattern
+            .compile("(([^/]+\\.[^/]+)/)?(([a-z0-9_]+)/)?([a-zA-Z0-9-_\\.]+)(:([a-z0-9-_\\.]+))?");
+
     /**
      * Null if this is on the public docker hub.
      */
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
```diff
@@ -53,6 +72,24 @@ public DockerRegistryEndpoint(String url, String credentialsId) {
         this.credentialsId = Util.fixEmpty(credentialsId);
     }
 
+    public static DockerRegistryEndpoint parse(String repo) {
```
jglick: Certainly needs Javadoc.

carlossg:
Added javadoc and also the missing parse of `:tag` just in case

src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
```diff
@@ -54,6 +71,34 @@ public DockerRegistryEndpoint(String url, String credentialsId) {
     }
 
     /**
+     * Parse the registry endpoint out of a registry:port/namespace/name:tag string. Credentials are set to the id
+     * passed.
+     * 
+     * @param s
+     * @param credentialsId
+     *            can be null
```
oleg-nenashev: Just add <code>@CheckForNull</code> annotation to the field

src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
```diff
@@ -54,6 +71,34 @@ public DockerRegistryEndpoint(String url, String credentialsId) {
...
+        Matcher matcher = DOCKER_REGISTRY_PATTERN.matcher(s);
+        if (!matcher.matches() || matcher.groupCount() < 7) {
+            throw new IllegalArgumentException(s + " does not match regex " + DOCKER_REGISTRY_PATTERN);
+        }
+        String url;
+        try {
+            // docker push always uses https
+            url = matcher.group(2) == null ? null : new URL("https://" + matcher.group(2)).toString();
+        } catch (MalformedURLException e) {
+            throw new IllegalArgumentException(s + " can not be parsed as URL: " + e.getMessage());
```
oleg-nenashev: IMO should be mentioned in Javadoc

jglick:
You have a compilation error here.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
carlossg: @oleg-nenashev do you mean @Nullable in the credentialsId param?

@CheckForNull is already set in the credentialsId field

oleg-nenashev: Sorry, it was a typo. The initial message was supposed to have: "...add <code>@CheckForNull</code> annotation to the **parameter**..."

carlossg:
damn eclipse, let's try again

oleg-nenashev:
:+1:

jglick:
BTW when you rebase commits inside the PR, I cannot see what you changed since the last time I looked, so I have to reread the whole thing.

src/test/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpointTest.java
```diff
@@ -0,0 +1,33 @@
+package org.jenkinsci.plugins.docker.commons;
```
jglick: Missing license header.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
```diff
@@ -54,6 +71,36 @@ public DockerRegistryEndpoint(String url, String credentialsId) {
     }
 
     /**
+     * Parse the registry endpoint out of a registry:port/namespace/name:tag string. Credentials are set to the id
+     * passed. The url is built from registry:port into https://registry:port, the same way docker push does.
+     * 
+     * @param s
+     * @param credentialsId
+     *            can be null
```
jglick: It was not obvious to me until I read the sources that this was not used as such, merely passed along to the `DockerRegistryEndpoint` constructor. Worth clarifying.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpoint.java
```diff
@@ -54,6 +71,36 @@ public DockerRegistryEndpoint(String url, String credentialsId) {
...
+     * @throws IllegalArgumentException
+     *             if string can't be parsed
+     * @return The DockerRegistryEndpoint corresponding to the registry:port part of the string
+     */
+    public static DockerRegistryEndpoint parse(String s, @CheckForNull String credentialsId) {
```
jglick: `parse` is a pretty generic name, and the input string here is not really a representation of the endpoint. It is an image name, as constructed by the `imageName` method. So I would suggest calling this `fromImageName`, and adding bidirectional Javadoc links with `imageName` to make it clear that these are related operations.

src/test/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpointTest.java
carlossg: not particularly obvious when only one file in the project has it. MIT? ASL?

jglick: MIT ought to be specified in the POM no?

carlossg:
Addressed comments. Should typically rebase before merging or up to me?

src/test/java/org/jenkinsci/plugins/docker/commons/DockerRegistryEndpointTest.java
```diff
@@ -0,0 +1,51 @@
+/*
+ * Copyright 2015 Carlos Sanchez <carlos@apache.org>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
```
jglick: The POM does specify MIT.

jglick:
:+1: other than incorrect license header.

Up to the merger to rebase or not. I never do.

oleg-nenashev:
I re-confirm :+1:

carlossg:
Merged in d0d19c6

PR #25, opened 2015-05-06
DockerImageExtractor extension point

rsandell:


rsandell:
@reviewbybees

rsandell:
@jglick

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+ * extension point to integrate with for example Docker HUB web hook support.
+ */
+public abstract class DockerImageExtractor implements ExtensionPoint {
+
+    /**
+     * Provides a collection of repository names {@code namespace/name} that the job uses as seen by the implementing class.
+     * Return an empty collection if none is found.
+     *
+     * @param job the job being queried.
+     * @return a collection of names, an or an empty collection.
```
jglick: stray `an`

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+     */
+    @Nonnull
+    public abstract Collection<String> getDockerImagesUsedByJob(Job<?,?> job);
```
jglick: Just for completeness, can mark `job` as `@Nonnull` too.

jglick:
:+1:

rsandell:
@jglick fixed

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+import jenkins.model.Jenkins;
+
+import javax.annotation.Nonnull;
+import java.util.Collection;
+
+/**
+ * Identifies the Docker images used by a Jenkins job. Docker-related plugins can use this
+ * extension point to integrate with for example Docker HUB web hook support.
+ */
+public abstract class DockerImageExtractor implements ExtensionPoint {
```
oleg-nenashev: What about <code>DockerImageIdContributor</code>?

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
oleg-nenashev: IMO we demand the same method for <code>Run</code>s

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * Provides a collection of repository names {@code namespace/name} that the job uses as seen by the implementing class.
+     * Return an empty collection if none is found.
+     *
+     * @param job the job being queried.
+     * @return a collection of names, or an empty collection.
+     */
+    @Nonnull
+    public abstract Collection<String> getDockerImagesUsedByJob(@Nonnull Job<?,?> job);
```
oleg-nenashev: What about <code>Run</code>s? // GitHub ate my previous comment

rsandell: There is no consumer of that that we know of currently, but it can be extended later if there is need.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
rsandell: Well, its not a contributor to me, it doesn't contribute to the job, it finds image names that has been configured in a job. I also didn't like the name at first glance, but haven't figured out one that explains it in a similar sense yet, so extractor remained.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public static ExtensionList<DockerImageExtractor> all() {
```
jglick: Instead of/in addition to making `all` public, I would suggest a `static` method to get the aggregate `Set<String>` for a given `Job` across all implementations.

rsandell:
@jglick Like that?

src/main/java/org/jenkinsci/plugins/docker/commons/DockerImageExtractor.java
```diff
@@ -0,0 +1,70 @@
...
+    /**
+     * Provides a set of repository names {@code namespace/name} that the job uses as seen by all the declared {@link DockerImageExtractor}s.
+     * Returns an empty set if none is found.
+     *
+     * @param job the job being queried.
+     * @return a set of names, or an empty set.
+     */
+    @Nonnull
+    public static Set<String> getDockerImagesUsedByJobFromAll(@Nonnull Job<?,?> job) {
+        Jenkins j = Jenkins.getInstance();
```
jglick: BTW the original `TODO` comment about `ExtensionList.lookup` was a useful reminder for when we upgrade the core dep; please resurrect it.

jglick:
:+1:

PR #26, opened 2015-05-06
Introduce DockerFingerprints::getImageFingerprintHash() method.

oleg-nenashev:
This public method is useful, because it allows to avoid a high-cost Fingerprint loading during hash calculations (e.g. links in web interfaces).

@reviewbybees


jglick:
:+1:

stephenc:
:+1:

PR #28, opened 2015-05-08
[FIXED JENKINS-28776] Add docker icons for actions.

oleg-nenashev:
These icons can be also shared across all other Docker plugins.

@reviewbybees


gusreiber:
:-1: These icons aren't really going to work. The small ones in particular, the 16x16 and 24x24 are way too small to have text in them. I'll see what I can figure out for you.

oleg-nenashev:
@gusreiber 
If there's no short-term icons proposal, let's merge the current ones to the alpha version. It will unblock the icons usage in other plugins. After that, you can just create a pull request replacing the icons by cool ones.

jglick:
Check with @tfennelly about `l:icon` usage.

oleg-nenashev:
Reassigning the task to @gusreiber . This PR is not a release blocker

oleg-nenashev:
@reviewbybees Changed the implementation to the icon class as it has been suggested by @jglick

src/main/java/org/jenkinsci/plugins/docker/commons/DockerCommonsPlugin.java
```diff
@@ -0,0 +1,49 @@
...
+import org.jenkins.ui.icon.IconType;
+
+/**
+ * Plugin class for Docker commons plugin.
+ * @author Oleg Nenashev
+ */
+public class DockerCommonsPlugin extends Plugin {
+    
+    @Override
+    public void start() throws Exception {
```
rsandell: This is an old pattern, you should use an ```@Initializer``` instead.

pom.xml
```diff
@@ -61,6 +61,11 @@
       <version>1.3</version>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.icon-shim</groupId>
+      <artifactId>icon-shim</artifactId>
```
rsandell: why not bump the core dependency to something that has this lib in it?

rsandell: Looking at the removed jelly code it seems like the correct core dependency is already there so you shouldn't need this plugin.

oleg-nenashev: No opinion - the plugin has been designed for this version, docker-traceability supports it as well

src/main/java/org/jenkinsci/plugins/docker/commons/DockerCommonsPlugin.java
ndeloof: agree. not a blocker but would prefer we keep good code samples as guidance for other plugins development

ndeloof:
:bee:

daniel-beck:
What's the source and license of these icons? Do we need to include a copyright/license statement somewhere? :bug: while unanswered.

oleg-nenashev:
We are eligible to use this icon until we don't modify it (e.g. remove the text).
From https://www.docker.com/trademark-guidelines...

> Use of Logos
The Docker’s registered, unregistered and pending logos, including the Docker “Moby Dock” whale logo, as well as other design elements such as characters and container images, may be taken only from Docker’s product sheet or from Docker’s service screen shot following authorization from us. Do not create imitations, facsimiles or reproductions of these design elements on your own. The design elements may only appear with the words, logotype, graphic elements and spacing which Docker has previously approved for use and may not be modified, recombined or displayed in modified scenes, images or other artwork.

daniel-beck:
@oleg-nenashev 

It looks like the name of the plugin may require permission… :frowning:  Still, out of scope of this PR.

However, it looks like we need to add a copyright statement somewhere (or should document this requirement for downstream users of icons). From that page:

> your use of Docker’s copyrighted material must be accompanied one of the following statements, which should be prominently displayed and reference the year/s the material was created:
> “Copyright 2013-2015 Docker, Inc. All rights reserved.”
> or
> “© 2013-2015 Docker, Inc. All rights reserved.”

src/main/java/org/jenkinsci/plugins/docker/commons/DockerCommonsPlugin.java
```diff
@@ -0,0 +1,49 @@
...
+        super.start();
+        registerIcons();
+    }
+    
+    private void registerIcons() {
+        IconSet.icons.addIcon(new Icon("icon-docker-logo icon-sm", "docker-commons/images/16x16/docker.png", Icon.ICON_SMALL_STYLE, IconType.PLUGIN));
```
jglick: Oof, really, yet another startup task? @tfennelly is there no on-demand loading of icons, say by looking for a resource in `uberClassLoader` of a given name?

oleg-nenashev: Could be an extension point BTW...

KostyaSha:
Status? Not mergeable

oleg-nenashev:
@daniel-beck Are you fine with the clarification from the Legal? See the corporate inbox (September?)

daniel-beck:
@oleg-nenashev Should be good enough. I still have some concerns about the icon size/legibility, but that's it.

oleg-nenashev:
Merging since there is no -1s left

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -45,7 +45,7 @@
  *
  * @author Oleg Nenashev
  */
-public class DockerFingerprintAction implements RunAction2 {
+public class DockerFingerprintAction implements RunAction2, org.jenkins.ui.icon.IconSpec {
```
KostyaSha: why import with package?

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -146,4 +146,8 @@ static void addToRun(Fingerprint fp, String imageId, Run run) throws IOException
             } // else no need to save updates
         }
     }
+
+    public String getIconClassName() {
```
KostyaSha: is it Override?

PR #29, opened 2015-05-08
Adding DockerTool

jglick:
For most purposes you can just use `docker` in `$PATH`. However there are some cases where this is not good enough:
   * The slave does not have Docker installed, and we want to be able to install it on demand.
   * The default installation is the wrong version to connect to a particular server. I have found that a 1.5.0 Swarm server rejects some (not all) commands coming from a 1.6.0 client.

- [X] define API
- [x] demonstrate its use in `SampleDockerBuilder`

@reviewbybees


oleg-nenashev:
:+1: for the feature. I was also going to implement it at some point.

Feature  proposal: It worths to inject environment variables before the client call. I suppose the following variables could be set up using the tool configuration.
* DOCKER_HOST
* DOCKER_TLS_VERIFY
* DOCKER_CERT_PATH

jglick:
Those are all set by `DockerServerEndpoint` materialization, regardless of client version.

oleg-nenashev:
I confirm :+1:

PR #30, opened 2015-05-10
DockerFingerprints API update

oleg-nenashev:
- [x] - Class can be also used for container fingerprints
- [x] - make() methods are accessible via API
- [x] - Extra helper methods, which allow to easily retrieve facets from Docker fingerprints

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
@@ -7,15 +7,23 @@ import jenkins.model.Jenkins;
 
 import java.io.IOException;
 import java.util.Collection;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.model.FingerprintFacet;
+import org.apache.commons.collections.CollectionUtils;
 
 /**
  * Entry point into fingerprint related functionalities in Docker.
  * This class provide basic methods for both images and containers
  */
 public class DockerFingerprints {
+    
+    private static final Logger LOGGER = Logger.getLogger(DockerFingerprints.class.getName());
+    
     private DockerFingerprints() {} // no instantiation
  
     /**
@@ -45,6 +53,15 @@ public class DockerFingerprints {
         return Jenkins.getInstance().getFingerprintMap().get(getFingerprintHash(id));
     }
     
+    private static @CheckForNull Fingerprint ofNoException(@Nonnull String id) {
+        try {
+            return of(id);
+        } catch (IOException ex) { // The error is not a hazard in CheckForNull logic
+            LOGGER.log(Level.WARNING, "Cannot retrieve a fingerprint for Docker id="+id, ex);
+        }
+        return null;
+    }
+    
     /**
      * Get or create a {@link Fingerprint} for the image.
      * @param run Origin of the fingerprint (if available)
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -7,50 +7,158 @@
...
     }
-
-    private static @Nonnull Fingerprint make(@Nonnull Run<?,?> run, @Nonnull String imageId) throws IOException {
-        return Jenkins.getInstance().getFingerprintMap().getOrCreate(run, "<docker-image>", getImageFingerprintHash(imageId));
+    
+    private static @CheckForNull Fingerprint ofNoException(@Nonnull String id) {
+        try {
+            return of(id);
+        } catch (IOException ex) { // The error is not a hazard in CheckForNull logic
+            LOGGER.log(Level.WARNING, "Cannot retrieve a fingerprint for Docker id="+id, ex);
```
jglick: Should this really be `WARNING`? Seems like it is fine to ask about a nonexistent ID, in which case this should be `FINE`.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -7,50 +7,158 @@
...
+    }
+    
+    /**
+     * Get or create a {@link Fingerprint} for the image.
+     * @param run Origin of the fingerprint (if available)
+     * @param id Image ID. Only 64-char full IDs are supported.
+     * @return Fingerprint for the specified ID
+     * @throws IOException Fingerprint load/save error
+     */
+    public static @Nonnull Fingerprint makeForImage(@CheckForNull Run<?,?> run, @Nonnull String id) throws IOException {
```
jglick: Would recommend just calling this `forImage`, reads better.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -7,50 +7,158 @@
...
+        return Jenkins.getInstance().getFingerprintMap().getOrCreate(run, "<docker-container>", getFingerprintHash(id));
     }
 
     /**
+     * Retrieves a facet from the {@link Fingerprint}.
+     * The method suppresses {@link IOException} if a fingerprint loading fails.
+     * @param <TFacet> Facet type to be retrieved
+     * @param id Docker item ID. Only 64-char full IDs are supported
+     * @param facetClass Class to be retrieved
+     * @return First matching facet. Null may be returned if the loading fails
```
jglick: Will also be null if there is such a fingerprint but no such facet, right?

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -7,50 +7,158 @@
...
+     */
+    @SuppressWarnings("unchecked")
+    public static @Nonnull <TFacet extends FingerprintFacet> Collection<TFacet> getFacets
+            (@Nonnull String id, @Nonnull Class<TFacet> facetClass) { 
+        final Fingerprint fp = ofNoException(id);
+        return (fp != null) ? getFacets(fp, facetClass) : CollectionUtils.EMPTY_COLLECTION;
```
jglick: `Collections.<TFacet>emptySet()`

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -7,50 +7,158 @@
...
+    }        
+    
+    //TODO: deprecate and use the core's method when it's available
```
jglick: So can these be `private` in the meantime?

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
oleg-nenashev: No. <code>IOException</code> may happen when loading an existing fingerprint from the disk. The method returns null for non-existent fingerprints

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
oleg-nenashev: yes. I'll fix javadoc

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
oleg-nenashev: The idea is to provide this method in API till Jenkins core supports it (I have not even created a PR yet)

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: (OK)

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: But it looks like it is only used internally in this class, right? So we could avoid polluting the `docker-commons` API with it?

oleg-nenashev: I use this methods in one of plugins under the development, so I would prefer to keep them public

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -7,50 +7,159 @@
...
+     * @param <TFacet> Facet type to be retrieved
+     * @param id Docker item ID. Only 64-char full IDs are supported
+     * @param facetClass Class to be retrieved
+     * @return First matching facet. Null may be returned if there is no facet
+     *      or if the loading fails
+     */
+    public static @CheckForNull @SuppressWarnings("unchecked")
+            <TFacet extends FingerprintFacet> TFacet getFacet
+            (@Nonnull String id, @Nonnull Class<TFacet> facetClass) { 
+        final Fingerprint fp = ofNoException(id);
```
tfennelly: NIT: the funky line wrapping here makes this a tease to read.

tfennelly:
:+1:

PR #31, opened 2015-05-12
Fixing DockerTool API to support ToolInstaller

jglick:
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
@@ -26,6 +26,7 @@ package org.jenkinsci.plugins.docker.commons;
 
 import hudson.EnvVars;
 import hudson.Extension;
+import hudson.FilePath;
 import hudson.Util;
 import hudson.model.EnvironmentSpecific;
 import hudson.model.Node;
@@ -83,6 +84,12 @@ public class DockerTool extends ToolInstallation implements EnvironmentSpecific<
                         }
                         String home = Util.fixEmpty(tool.getHome());
                         if (home != null) {
+                            if (node != null) {
+                                FilePath homeFP = node.createPath(home);
+                                if (homeFP != null) {
+                                    return homeFP.child("bin/docker").getRemote();
+                                }
+                            }
                             return home + "/bin/docker";
                         }
                     }
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
@@ -26,7 +26,6 @@ package org.jenkinsci.plugins.docker.commons;
 
 import hudson.EnvVars;
 import hudson.Extension;
-import hudson.Launcher;
 import hudson.Util;
 import hudson.model.EnvironmentSpecific;
 import hudson.model.Node;
@@ -37,14 +36,13 @@ import hudson.tools.ToolInstallation;
 import hudson.tools.ToolInstaller;
 import hudson.tools.ToolProperty;
 import hudson.util.ArgumentListBuilder;
-import hudson.util.FormValidation;
 import java.io.IOException;
 import java.util.List;
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import jenkins.model.Jenkins;
 import org.kohsuke.stapler.DataBoundConstructor;
-import org.kohsuke.stapler.QueryParameter;
 
 /**
  * An installation of Docker.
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
```diff
@@ -66,16 +64,27 @@
...
             if (j != null) {
                 for (DockerTool tool : j.getDescriptorByType(DescriptorImpl.class).getInstallations()) {
                     if (tool.getName().equals(name)) {
-                        return tool.getExecutable(launcher);
+                        if (node != null) {
+                            tool = tool.forNode(node, listener);
+                        }
+                        if (env != null) {
+                            tool = tool.forEnvironment(env);
+                        }
```
tfennelly: Is the above any different to:

```
if (env != null) {
    tool = tool.forEnvironment(env);
} else if (node != null) {
    tool = tool.forNode(node, listener);
}
```

Aside from the fact that the above only flows through one leg of the `if` and (imo) more clearly shows that `env` takes precedence over `node`.

ndeloof:
:+1:

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
```diff
@@ -66,16 +64,27 @@
...
+                        String home = Util.fixEmpty(tool.getHome());
+                        if (home != null) {
+                            return home + "/bin/docker";
```
oleg-nenashev: Hardcoded platform-dependent slashes. <code>getHome()</code> may return Windows path from the installer, so it cannot be considered as a standard path.

oleg-nenashev:
LGTM (excepting the platform compatibility issue)

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
tfennelly: Can you guarantee this? I know OSX might not be so important, but boot2docker puts `docker` in `/usr/local/bin/` and doesn't put anything in `~/bin`.

ndeloof: If this DockerTool is used for client CLI, then need to support windows (yes, Docker released a windows native docker client).

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
jglick: Yes, you want to allow _both_ `NodeSpecific` and `EnvironmentSpecific` to apply. This is simply inlining [`ToolInstallation.translate`](https://github.com/jglick/jenkins/blob/caedb49c23f2ec24dae9c28f8a604864dc1fbb3e/core/src/main/java/hudson/tools/ToolInstallation.java#L170-177) without requiring a cast back to `DockerTool` at the end.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
jglick: > `getHome()` may return Windows path from the installer

If you had a Docker CLI installed on Windows, yes it could. The code would likely still work because Windows would accept an executable path with mixed `\` and `/`. It could be changed to pick up the file separator from the `node` (possibly at the expense of a `Callable` roundtrip) if you think this is a real problem.

jglick: > boot2docker puts `docker` in `/usr/local/bin/`

Fine, then `home` would be `/usr/local`.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
tfennelly: Ah ok ... there is some sort of overlaying going on here. I think a comment to that effect would be good otherwise most people (like me) don't know something funky is happening.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
tfennelly: Ah ok ... thanks ... `home` in this case is `toolHome` as opposed to the home dir of the current user.

tfennelly:
:+1:

jglick:
Tweaked the code to work with a Windows slave. Tested using a Docker CLI installation on XP (with a Linux master); `getExecutable` as run from the script console uses the correct `\` separator.

oleg-nenashev:
:+1:

ndeloof:
:+1:

tfennelly:
:+1:

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
```diff
@@ -66,16 +65,33 @@
...
+                        }
+                        if (env != null) {
+                            tool = tool.forEnvironment(env);
+                        }
+                        String home = Util.fixEmpty(tool.getHome());
+                        if (home != null) {
+                            if (node != null) {
+                                FilePath homeFP = node.createPath(home);
+                                if (homeFP != null) {
+                                    return homeFP.child("bin/docker").getRemote();
```
tfennelly: Is there a chance of doing an `exists` here and logging an error or warning if `docker` doesn't exist? I'm sure it would become obvious after when things fail, but at least you could warn with some info wrt where we expect `docker` to be.

Maybe the whole Tooling infra makes sure this can't happen.

jglick: Might be useful to add. `ToolInstallation` does not guarantee the file exists. Of course you would get _some_ kind of error from the build anyway, but might be clearer if that were printed right here (assuming `listener != null`).

Anyway too late for this PR.

src/main/java/org/jenkinsci/plugins/docker/commons/DockerTool.java
jglick: > there is some sort of overlaying going on here

Well, the tool can be node-specific (generally due to using a `ToolInstaller`, but also a `ToolLocationNodeProperty`). More rarely it can expand environment variables in `home`. They are orthogonal. This is just a standard pattern for any use of a `ToolInstallation`.

PR #32, opened 2015-05-14
Docker Facets Base class + summaries in the UI

oleg-nenashev:
Introduces DockerFingerprintFacet base class, which may contribute the UI. The current implementation allows to add summaries to DockerFingerprintAction. It allows to make it much more informative.

@reviewbybees

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction/index.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction/index.jelly
@@ -34,7 +34,7 @@ THE SOFTWARE.
       <table class="sortable bigtable pane">
         <thead style="text-align:left">
           <tr>
-            <th>${%Image ID}</th>
+            <th>${%Image Summary}</th>
             <th>${%Created on}</th>
             <th>${%Introduced}</th>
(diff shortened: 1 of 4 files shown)


tfennelly:
:+1:

src/main/resources/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction/index.jelly
```diff
@@ -45,13 +45,18 @@ THE SOFTWARE.
...
                   <td>${%Unknown}</td>
                   <td>${%outside Jenkins}</td>     
                 </j:when>
                 <j:otherwise>
                   <td>
-                    <a href="${rootURL}/fingerprint/${fp.hashString}">${imageId}</a>
+                    ${%ID}: <a href="${rootURL}/fingerprint/${fp.hashString}">${imageId}</a>
+                    <j:forEach var="dockerFacet" items="${it.getDockerFacets(imageId)}">
+                      <div>
+                        <st:include page="summary.jelly" from="${dockerFacet}" optional="true" it="${dockerFacet}" />
```
jglick: Can probably remove the `from` attribute; IIRC it is defaulted to `it`.

jglick:
:+1:

PR #34, opened 2015-05-20
hide Docker installation option if no installation is defined

ndeloof:
For users (most of them imho) who will just rely on docker installation in default PATH

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/docker/commons/lib/selectDockerTool.jelly
+++ b/src/main/resources/org/jenkinsci/plugins/docker/commons/lib/selectDockerTool.jelly
@@ -33,12 +33,15 @@ THE SOFTWARE.
         </st:attribute>
     </st:documentation>
     <!-- TODO this is pretty generic, perhaps it should be generalized to a tag in Jenkins core -->
-    <f:entry title="${%Docker installation}">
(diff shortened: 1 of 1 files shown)


ndeloof:
@reviewbybees

oleg-nenashev:
:+1:

oleg-nenashev:
Seems there's a merge conflict

ndeloof:
I just rebased

jtnord:
:+1:

jglick:
:-1:, this contradicts [JENKINS-17373](https://issues.jenkins-ci.org/browse/JENKINS-17373) and I think it should be reverted.

ndeloof:
@jesse proposed fix for JENKINS-17373 is the worst I can imagine : make the UI bad to prevent misconfigurations. Jenkins should just prevent to remove a ToolInstallation being used by some job and propose to rename them (as it does for job references).

jglick:
> Jenkins should just prevent to remove a ToolInstallation being used by some job

Can try to do so in certain cases, but incompatible with code-as-config.

ndeloof:
code as configuration would let you say "I need java 8" or "I need open JDK 8u45", then it's jenkins job to determine which ToolInstallation do match this vendor/version requirements. Relying on arbitrary display name is fragile and not portable across jenkins instances.

PR #35, opened 2015-05-21
Move docker-commons taglib to /lib/docker/commons

oleg-nenashev:
As @ndeloof suggested

@reviewbybees


src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerTool.java
```diff
@@ -48,7 +48,7 @@
 /**
  * An installation of Docker.
  * If nonempty, expected to have an executable file {@code bin/docker}.
- * Use {@code <d:selectDockerTool field="toolName" xmlns:d="/org/jenkinsci/plugins/docker/commons/lib"/>} to customize.
+ * Use {@code <d:selectDockerTool field="toolName" xmlns:d="/lib/docker/commons/"/>} to customize.
```
ndeloof: no need for a trailing `/` in namespace afaik

ndeloof:
:+1:

tfennelly:
:+1:

jglick:
While it is too late now, FTR: I intentionally used the same package namespace as the plugin itself. IMO only core should be able to use the `/lib` namespace. But the only precedent that I know of is in Credentials.

oleg-nenashev:
@jglick 
I have no strong opinion on this change. We could deprecate the existing taglib and create it's copy at the right location.

PR #36, opened 2015-06-02
Integrate docker-java dependency.

oleg-nenashev:
It's a follow-up to the discussion in https://github.com/jenkinsci/docker-build-publish-plugin/pull/22/files#r31504279.
We use docker-java in several docker plugins, so it makes sense to streamline the dependency via api plugin like docker-commons.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -51,6 +51,11 @@
       <version>1.22</version>
     </dependency>
     <dependency>
+      <groupId>com.github.docker-java</groupId>
+      <artifactId>docker-java</artifactId>
+      <version>1.3.0</version>
+    </dependency>
+    <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>authentication-tokens</artifactId>
       <version>1.1</version>


ndeloof:
:+1:

jglick:
-0. I think it is safer to bundle this in its own plugin, especially as I have heard that they are not very strict about maintaining backward compatibility—in which case we would need a `docker-java-13` plugin, a `docker-java-14` plugin, etc.

oleg-nenashev:
> -0. I think it is safer to bundle this in its own plugin, especially as I have heard that they are not very strict about maintaining backward compatibility—in which case we would need a docker-java-13 plugin, a docker-java-14 plugin, etc.

Technically we would need such wrapper plugin for each lib we use in Jenkins. I'm OK to create an additional wrapper, but it causes extra overheads.

ndeloof:
I'm fine with a dedicated docker-java plugin (I plan to do the same with git-client and split JGit in it's own plugin so it can burn in hell), but if the goal is to prevent backward compatibility issues, then it would make more sense imho to have the adequate docker-java lib bundled with each plugin, better than a dozen docker-java-<version> plugins installed

jglick:
Keeping the library bundled in each plugin is certainly safer, albeit at the cost of duplicating code. Sharing a wrapper plugin is better when you are reasonably confident that the authors pay some attention to maintaining backward compatibility.

oleg-nenashev:
> Sharing a wrapper plugin is better when you are reasonably confident that the authors pay some attention to maintaining backward compatibility.

Investigated docker-java commit history. There's no backward compatibility at all, hence I'll fall-back to shading

PR #37, opened 2015-06-05
[FIXED JENKINS-28657] - Support the creation of named fingerprints

oleg-nenashev:
The fix also changes the default behavior for missing names. Instead of non-informative "<docker-image>" we will be using "Docker image $(64-char id)" (which may be too long BTW)

@reviewbybees


jglick:
[JENKINS-28657](https://issues.jenkins-ci.org/issues/JENKINS-28657)

Abstaining until there is a downstream PR in some other plugin demonstrating how this would be useful.

oleg-nenashev:
See the latest Docker traceability demo BTW (it uses its own named fingerprints now). Anyway, the patch can be postponed

ndeloof:
:bee:

PR #38, opened 2015-06-28
Support Docker 1.7+ .docker/config.json

ndeloof:
Docker 1.7 introduced `.docker/config.json` which offer configuration extensibility. This file is preferred over `.dockercfg`, so if docker node has already ran some docker stuff with 1.7, jenkins just can't set credentials based on `.dockercfg`


ndeloof:
@reviewbybees

carlossg:
LGTM +1

michaelneale:
:100:

jglick:
BTW if https://github.com/docker/docker/pull/12856 and/or https://github.com/docker/docker/issues/10318 were addressed, the logic here could be improved (insofar as I understand those proposals).

oleg-nenashev:
Caused https://issues.jenkins-ci.org/browse/JENKINS-29627, which has been fixed in docker-commons-1.2

PR #39, opened 2015-06-29
option to configure a default dockerhost value

ndeloof:
customizing dockerhost is an advanced usage for complex deployment workflows, and I guess most users will just rely on the same dockerhost for all jobs (especially considering docker-swarm usage).
BUT in some setup docker local socket is not set as /var/run/docker.sock

This PR allows administrator to configure a default dockerhost URL to be used when not overiden explicitly on job.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
@@ -40,12 +40,16 @@ import hudson.remoting.VirtualChannel;
 import hudson.util.ListBoxModel;
 import jenkins.authentication.tokens.api.AuthenticationTokens;
 import jenkins.model.Jenkins;
+import net.sf.json.JSONObject;
+import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.docker.commons.impl.ServerHostKeyMaterialFactory;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 import org.kohsuke.stapler.QueryParameter;
+import org.kohsuke.stapler.StaplerRequest;
 
 import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import java.io.IOException;
 import java.util.List;
@@ -78,9 +82,18 @@ public class DockerServerEndpoint extends AbstractDescribableImpl<DockerServerEn
      *
      * <p>
      * Null to indicate whatever Docker picks by default.
+     * @deprecated use #getDockerHost
      */
     public @Nullable String getUri() {
         return uri;
+    }
+
+    /**
+     * Gets the endpoint in URI, such as "unix:///var/run/docker.sock".
+     */
+    public @Nonnull String getDockerHost() {
+        if (StringUtils.isNotBlank(uri)) return uri;
+        return getDescriptor().getDefaultDockerHost();
     }
 
     /**
@@ -166,15 +179,51 @@ public class DockerServerEndpoint extends AbstractDescribableImpl<DockerServerEn
         }
         return true;
     }
-    
+
+
+    @Override
+    public DescriptorImpl getDescriptor() {
+        return (DescriptorImpl) super.getDescriptor();
+    }
+
     @Extension
     public static class DescriptorImpl extends Descriptor<DockerServerEndpoint> {
+
+        private String defaultHost;
+
+        public String getDefaultHost() {
+            return defaultHost;
+        }
+
+        public DescriptorImpl() {
+            load();
+        }
+
+        @Override
+        public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
+            defaultHost = json.getString("defaultHost");
+            save();
+            return true;
+        }
+
+        public String getDefaultDockerHost() {
+            if (StringUtils.isNotBlank(defaultHost))
+                return defaultHost;
+            String env = System.getenv("DOCKER_HOST");
+            if (env != null)
+                return env;
+            return "unix:///var/run/docker.sock";
+        }
+
         @Override
         public String getDisplayName() {
             return "Docker Daemon";
         }
 
         public ListBoxModel doFillCredentialsIdItems(@AncestorInPath Item item, @QueryParameter String uri) {
+
+            if (StringUtils.isBlank(uri)) uri = getDefaultDockerHost();
+
             List<DomainRequirement> domainRequirements = URIRequirementBuilder.fromUri(uri).build();
             domainRequirements.add(new DockerServerDomainRequirement());
             return new StandardListBoxModel()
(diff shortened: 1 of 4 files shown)


ndeloof:
@reviewbybee

jtnord:
@ndeloof I think you mean @reviewbybees

amuniz:
If this PR is related an existent JIRA task, it should be indicated in the title (as [JENKINS-XXXX]) and commit messages.

ndeloof:
there's no associated Jira

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
```diff
@@ -78,12 +82,21 @@ public DockerServerEndpoint(String uri, String credentialsId) {
...
+     * @deprecated use #getDockerHost
      */
     public @Nullable String getUri() {
         return uri;
     }
 
     /**
+     * Gets the endpoint in URI, such as "unix:///var/run/docker.sock".
+     */
+    public @Nonnull String getDockerHost() {
```
jtnord: this is not a host but a URI/URL

cyrille-leclerc:
:bee:

src/main/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint/global.jelly
```diff
@@ -0,0 +1,31 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+The MIT License
+
+Copyright 2015 Jesse Glick.
```
jtnord: Who what....

src/main/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint/global.jelly
```diff
@@ -0,0 +1,31 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="defaultHost" title="Default Docker host URL">
```
jtnord: its not a host.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
ndeloof: I used "dockerhost" to map the DOCKER_HOST environment variable used for CLI usage

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
```diff
@@ -166,15 +179,51 @@ public boolean equals(Object obj) {
...
+        }
+
+        @Override
+        public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
+            defaultHost = json.getString("defaultHost");
+            save();
+            return true;
+        }
+
+        public String getDefaultDockerHost() {
```
jtnord: IMHO the default should be unset which would mean users need to get work as they currently do.

ndeloof: currently user don't need to do anything, as long as his ENV has DOCKER_HOST set. 
UX to force user to enter docker host URL to all jobs would be just bad

jtnord:
:bug: too may small questions for me so
I would like to see how you expect this to be used in existing plugins as well.

ndeloof:
@jtnord https://github.com/jenkinsci/docker-custom-build-environment-plugin/tree/dockerhost

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
```diff
@@ -78,12 +82,21 @@ public DockerServerEndpoint(String uri, String credentialsId) {
      *
      * <p>
      * Null to indicate whatever Docker picks by default.
+     * @deprecated use #getDockerHost
```
jglick: Do not deprecate this, it is part of the GUI configuration (getter corresponding to constructor parameter).

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
```diff
@@ -166,15 +179,51 @@ public boolean equals(Object obj) {
...
+        public boolean configure(StaplerRequest req, JSONObject json) throws FormException {
+            defaultHost = json.getString("defaultHost");
+            save();
+            return true;
+        }
+
+        public String getDefaultDockerHost() {
+            if (StringUtils.isNotBlank(defaultHost))
+                return defaultHost;
+            String env = System.getenv("DOCKER_HOST");
```
jglick: This is an environment variable on master and so is worse than useless as a default. :bug:

jglick:
> https://github.com/jenkinsci/docker-custom-build-environment-plugin/tree/dockerhost

BTW when defining a new API I strongly recommend creating at least one `[WiP]` PR for the proposed client calls and linking it to the API PR, so reviewers can see everything in context.

jglick:
IIUC this PR directly contradicts the recommendation @kohsuke originally made when creating this plugin, that server/registry endpoints be explicitly configured in jobs rather than using global settings (and that if users feel a need to share configuration between jobs, they do so using some orthogonal technique).

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
jglick: Agree with @jtnord that the naming is inappropriate. The existing naming is `DockerServerEndpoint.uri`. If you are defining a default value for this property, then the corresponding method would be something like `DockerServerEndpoint.getEffectiveUri`. Compare the existing `DockerRegistryEndpoint.getEffectiveUrl`.

jglick: I also agree with @jtnord that if no systemwide default has been configured, and the endpoint specifies no URI, it would be better to fall back to `null` meaning to not specify `--host` at all and use the default server according to the client on whatever machine it runs on.

ndeloof:
:'( 

and so user do complain about Jenkins UX ...

jglick:
To be clear, I am not personally voting against this, I am just saying that it conflicts with my recollection of a design decision made a couple of months ago. That decision may no longer be defensible, or I may be misstating or exaggerating it. It would be best if @kohsuke could weigh in here.

stephenc:
> It would be best if @kohsuke could weigh in here.

I agree

PR #41, opened 2015-06-30
Docker Tool Installer

ndeloof:
Download docker executable from docker.io and install automatically on slave
Please note: this do install docker cli for client operation, can't be used to run daemon which require extra setup.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
@@ -39,13 +39,14 @@ import org.kohsuke.stapler.DataBoundConstructor;
 
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
 import java.net.HttpURLConnection;
 import java.net.URL;
 import java.net.URLConnection;
 
 
 /**
- * Download and install Docker CLI binary, see // see https://docs.docker.com/installation/binaries/
+ * Download and install Docker CLI binary, see https://docs.docker.com/installation/binaries/
  * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
  */
 public class DockerToolInstaller extends ToolInstaller {
@@ -82,13 +83,19 @@ public class DockerToolInstaller extends ToolInstaller {
         // (simplified) copy/paste from FilePath as hudson.FilePath.installIfNecessaryFrom do assume the URL points to be a zip/tar archive
 
         FilePath timestamp = install.child(".timestamp");
-        URLConnection con;
+        URLConnection con = null;
+        long sourceTimestamp;
         try {
             con = ProxyConfiguration.open(url);
             if (timestamp.exists()) {
                 con.setIfModifiedSince(timestamp.lastModified());
             }
             con.connect();
+            if (con instanceof HttpURLConnection
+                    && ((HttpURLConnection)con).getResponseCode() == HttpURLConnection.HTTP_NOT_MODIFIED) {
+                return install;
+            }
+            sourceTimestamp = con.getLastModified();
 
         } catch (IOException x) {
             if (install.exists()) {
@@ -100,14 +107,10 @@ public class DockerToolInstaller extends ToolInstaller {
             } else {
                 throw x;
             }
+        } finally {
+            if (con instanceof HttpURLConnection) ((HttpURLConnection) con).disconnect();
         }
 
-        if (con instanceof HttpURLConnection
-                && ((HttpURLConnection)con).getResponseCode() == HttpURLConnection.HTTP_NOT_MODIFIED) {
-            return install;
-        }
-
-        long sourceTimestamp = con.getLastModified();
         if (install.exists()) {
             if (timestamp.exists() && sourceTimestamp == timestamp.lastModified())
                 return install;   // already up to date
@@ -124,16 +127,22 @@ public class DockerToolInstaller extends ToolInstaller {
             try {
                 install.act(new FilePath.FileCallable<Object>() {
                     public Object invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
-                        IOUtils.copy(url.openStream(), f);
+                        InputStream in = url.openStream();
+                        IOUtils.copy(in, f);
+                        in.close();
                         return null;
                     }
                 });
             } catch (IOException x) {
                 x.printStackTrace(listener.error("Failed to download " + url + " from slave; will retry from master"));
-                docker.copyFrom(ProxyConfiguration.getInputStream(url));
+                InputStream in = ProxyConfiguration.getInputStream(url);
+                docker.copyFrom(in);
+                in.close();
             }
         } else {
-            docker.copyFrom(ProxyConfiguration.getInputStream(url));
+            InputStream in = ProxyConfiguration.getInputStream(url);
+            docker.copyFrom(in);
+            in.close();
         }
 
         docker.act(new FilePath.FileCallable<Object>() {
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
@@ -1,0 +1,142 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick, CloudBees Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.jenkinsci.plugins.docker.commons.tools;
+
+import hudson.Extension;
+import hudson.FilePath;
+import hudson.ProxyConfiguration;
+import hudson.model.Node;
+import hudson.model.TaskListener;
+import hudson.remoting.Callable;
+import hudson.remoting.VirtualChannel;
+import hudson.tools.ToolInstallation;
+import hudson.tools.ToolInstaller;
+import hudson.tools.ToolInstallerDescriptor;
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import java.io.File;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLConnection;
+
+
+/**
+ * Download and install Docker CLI binary, see // see https://docs.docker.com/installation/binaries/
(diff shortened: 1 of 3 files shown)


ndeloof:
@reviewbybees

src/main/java/org/jenkinsci/plugins/docker/commons/Main.java
```diff
@@ -0,0 +1,11 @@
+package org.jenkinsci.plugins.docker.commons;
+
+/**
+ * @author <a href="mailto:nicolas.deloof@gmail.com">Nicolas De Loof</a>
+ */
+public class Main {
```
amuniz: What's the goal of this class?

jglick: Yeah delete please :bug:.

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,118 @@
+package org.jenkinsci.plugins.docker.commons.tools;
```
jglick: :bug: missing license header.

jglick:
This PR should delete [this comment](https://github.com/jenkinsci/docker-commons-plugin/blob/e6dd443c504ee2c81317341019392902dec155e3/src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerTool.java#L134).

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,118 @@
...
+        if (install.exists()) {
+            if(timestamp.exists() && sourceTimestamp ==timestamp.lastModified())
+                return install;   // already up to date
+            install.deleteContents();
+        }
+
+        FilePath bin = install.child("bin");
+        bin.mkdirs();
+        FilePath docker = bin.child("docker");
+        docker.copyFrom(ProxyConfiguration.getInputStream(url));
```
jglick: This will work, though it is unnecessarily slow in typical cases, since it opens the URL on the master, and streams the content over the slave channel. For regular tool installations using `installIfNecessaryFrom` we first try to perform the download from the slave, falling back to the master if necessary (say, if the slave has no proxy configuration, or no Internet connection). Perhaps the Docker CLI is small enough that we do not care about this issue, but FYI.

jglick: (This sort of subtlety is why a variant of `installIfNecessaryFrom` not assuming an archive would be a welcome addition to `FilePath`.)

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,118 @@
...
+        docker.act(new FilePath.FileCallable<Object>() {
+            public Object invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
+                f.setExecutable(true);
```
jglick: This is Java 6+. Are we not enforcing Animal Sniffer here? Maybe the OSS plugin parent POM omits that. Last I checked it was enforcing Java 5-. (Yes we want to assume 6+, I was just surprised that this compiles as is.)

src/main/resources/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller/config.jelly
```diff
@@ -0,0 +1,29 @@
...
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+-->
+<?jelly escape-by-default='true'?>
+<j:jelly xmlns:j="jelly:core" xmlns:f="/lib/form">
+    <f:entry field="version" title="Docker version">
+        <f:textbox default="latest"/>
```
jglick: Note that as a future RFE you could create a `Downloadable` paired with a server-side indexing job to offer completion here.

jglick:
Looks good other than the minor inline comments, thanks for implementing this!

src/main/java/org/jenkinsci/plugins/docker/commons/Main.java
ndeloof: oups

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
ndeloof: Marked as a :bug: for a license header, are you kidding ?
project has a LICENSE.txt file on scm root, why do we need such a license header ?
Do you add license headers to icon files as well ? :-P

src/main/resources/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller/config.jelly
ndeloof: Was my initial plan, but https://get.docker.com/builds/ is not indexable

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
ndeloof: Ah ok, I didn't got this reading installIfNecessaryFrom first time
Agree this could be a nice addition to FilePath, but need this running on current jenkins for my immediate needs. Maybe will consider as another core PR

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
ndeloof: So you mean OSS parent require an update to reflect switch to Java 6 which occurred ... 2 years ago ?

ndeloof:
@jglick 4385be157302113370091d073b36c11a93d65631 introduces download from slave then fallback to master as suggested (docker binary is ~ 16Mb)

ndeloof:
need to stop relying IDE to manage commits and switch back to CLI+GitX :-\

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,159 @@
+/*
+ * The MIT License
+ *
+ * Copyright 2015 Jesse Glick, CloudBees Inc.
```
jglick: Really? :-)

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
jglick: IANAL but my impression was that at least all `*.java` should have it. Anyway, resolved.

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
ndeloof: Copy past is my best friend !
Makes me wonder : License headers for CloudBees contributed code should not include committer name in Copyright, just Company name, don't you think so ?

jglick: IIUC our employer is intending to deliver guidelines on the topic so it need not be discussed here.

src/main/resources/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller/config.jelly
jglick: Ah. Anyway I find the feature in other installers to be of questionable value because the result can be incomplete or stale and the pulldown is not editable.

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,159 @@
...
+
+import java.io.File;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLConnection;
+
+
+/**
+ * Download and install Docker CLI binary, see // see https://docs.docker.com/installation/binaries/
```
jglick: Probably mistaken `// see`

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,159 @@
...
+        // (simplified) copy/paste from FilePath as hudson.FilePath.installIfNecessaryFrom do assume the URL points to be a zip/tar archive
+
+        FilePath timestamp = install.child(".timestamp");
+        URLConnection con;
+        try {
+            con = ProxyConfiguration.open(url);
+            if (timestamp.exists()) {
+                con.setIfModifiedSince(timestamp.lastModified());
+            }
+            con.connect();
```
jglick: Does this leak a socket if it is not closed?

jglick: :bug: I believe you need to call `((HttpURLConnection)con).disconnect()` in a `finally` block at the end of this whole method. You may also need to call `InputStream.close` around here, since either `connect` or `getResponseCode` may open the stream; and again below in three places, since `IOUtils.copy` does not close its input stream, and `FilePath.copyFrom(InputStream)` does not seem to either.

ndeloof: That's probably true. Please note then FilePath#installIfNecessaryFrom has the same issue

oleg-nenashev:
@ndeloof Why did you merge it without waiting for reviews?

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,168 @@
...
+            if (con instanceof HttpURLConnection
+                    && ((HttpURLConnection)con).getResponseCode() == HttpURLConnection.HTTP_NOT_MODIFIED) {
+                return install;
+            }
+            sourceTimestamp = con.getLastModified();
+
+        } catch (IOException x) {
+            if (install.exists()) {
+                // Cannot connect now, so assume whatever was last unpacked is still OK.
+                if (listener != null) {
```
oleg-nenashev: AFAIK listener is never null

ndeloof:
@oleg-nenashev "At least one :bee: comments from a CloudBees employee and at least 12 hours since the last commit.". code review is not about making things perfect before merge (as a such postpone them for months) but ensure another pair of eyes did check one's work to detect obvious bugs (and hopefully other ones as well if possible)

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -0,0 +1,168 @@
...
+        return version;
+    }
+
+    @Override
+    public FilePath performInstallation(ToolInstallation toolInstallation, Node node, TaskListener listener) throws IOException, InterruptedException {
+
+        String os = node.getChannel().call(new Callable<String, IOException>() {
+            public String call() throws IOException {
+                String os = System.getProperty("os.name").toLowerCase();
+                String arch = System.getProperty("os.arch").contains("64") ? "x86_64" : "i386";
```
oleg-nenashev: This calculation is incorrect. OS can be easily built for i586, etc.

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
ndeloof: this is the way it is implemented in FilePath#installIfNecessaryFrom 
as noticed, this is mostly a copy/paste due to a jenkins-core limitation

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
ndeloof: samples I've seen so far do report x86_64 or amd64.
This could be improved later, especially to support arm targets

oleg-nenashev:
:bug: till the ToolInstaller has a <code>help.html</code> describing known limitations. This file would be useful in any case.

oleg-nenashev:
> "At least one :bee: comments from a CloudBees employee and at least 12 hours since the last commit.".

Yes, this is a policy. But I don't see any :bee:. Jesse's LGTM has been overridden by :bug: later + there were other significant changes after such feedback (especially https://github.com/ndeloof/docker-commons-plugin/commit/75358d0828b642a9f51ba8f99d0c79786c941a03).

PR #42, opened 2015-07-02
prefer « Docker Host » terminology

ndeloof:
as this is what is used by Docker as well as env variable used by docker user to point to a docker daemon 
and we called « DockerServerEndpoint » in our own vocable

### DIFF
--- a/src/main/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint/help-uri.html
+++ b/src/main/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint/help-uri.html
@@ -1,5 +1,5 @@
 <div>
-    URI to the Docker server you are using.
-    May be left blank to use the Docker default
-    (typically <code>unix:///var/run/docker.sock</code>).
+    URI to the Docker Host you are using.
+    May be left blank to use the Docker default (defined by DOCKER_HOST environment variable)
+    (typically <code>unix:///var/run/docker.sock</code> or <code>tcp://127.0.0.1:2375</code>).
 </div>
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
@@ -96,7 +96,7 @@ public class DockerServerCredentials extends BaseStandardCredentials {
     public static class DescriptorImpl extends BaseStandardCredentialsDescriptor {
         @Override
         public String getDisplayName() {
-            return "Docker Server Certificate Authentication";
+            return "Docker Host Certificate Authentication";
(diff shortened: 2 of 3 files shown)


ndeloof:
@reviewbybees

src/main/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint/help-uri.html
```diff
@@ -1,5 +1,5 @@
 <div>
-    URI to the Docker server you are using.
-    May be left blank to use the Docker default
-    (typically <code>unix:///var/run/docker.sock</code>).
+    URI to the Docker Host you are using.
+    May be left blank to use the Docker default (defined by DOCKER_HOST environment variable)
+    (typically <code>unix:///var/run/docker.sock</code> or <code>tcp://127.0.0.1:2376</code>).
```
oleg-nenashev: NIT: It's better to mention 2375, because 2376 is being used for TLS-enabled hosts

ndeloof: I would not recommend anybody to run docker without TLS enabled ! :-)

oleg-nenashev: Mee too. BTW it's not our business to explain all other env. vars required to run on this port

oleg-nenashev:
Both "Docker Server" and "Docker Host" seem to be incorrect. In official docs they call it just "Docker", which seems to be confusing. Personally I'd put a :bee: on this change

src/main/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint/help-uri.html
ndeloof: But it's our job to make UX as smooth as possible and provide useful help 
changed port in 6537090

ndeloof:
Which official doc do you refer to ? 
https://docs.docker.com/swarm/ to just take a sample talks about "poll of _Docker Hosts_"

oleg-nenashev:
In https://docs.docker.com/userguide/usingdocker/ Docker Host is being references as Docker. BTW, Docker Machine also refers "Docker Host"

ndeloof:
This doc does not make reference to docker hosts but container management, so "Docker" in this context is equivalent to "Docker container", not related to Docker host

ydubreuil:
Standard documentation https://docs.docker.com/misc/ refers to Docker hosts when talking about multiple docker hosts managed by Docker machine and aligning with environment variable naming is a good idea, so :bee: for me

oleg-nenashev:
I confirm my :bee:

PR #45, opened 2015-07-23
[FIXED-JENKINS-29627] - Fix issue with new docker version

shinji62:
Hi,

I found a bug in the source code, when you install the plugin and docker 1.7.0 for the first time, it will create a config.json file which is fine for the new version but the Json format will use the old version which doesn't work with the 1.7.0

So by default for backward compatibilty, I create the old version in the old file and docker event the new version will be able to use it.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryToken.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryToken.java
@@ -86,24 +86,21 @@ public final class DockerRegistryToken implements Serializable {
                         json = JSONObject.fromObject(FileUtils.readFileToString(config, "UTF-8"));
                         auths = json.getJSONObject("auths");
                     } else {
-                        File dockercfg = new File(System.getProperty("user.home"), ".dockercfg");
-                        if (dockercfg.exists()) {
-                            config = dockercfg;
-                            auths = json = JSONObject.fromObject(FileUtils.readFileToString(dockercfg, "UTF-8"));
+                        config = new File(System.getProperty("user.home"), ".dockercfg");
+                        if (config.exists()) {
+                            auths = json = JSONObject.fromObject(FileUtils.readFileToString(config, "UTF-8"));
                         } else {
                             // Use legacy .dockercfg to ensure this works well with pre-1.7 docker client
                             // client will pick this one if .docker/config.json does not yet exists
                             auths = json = new JSONObject();
                         }
                     }
-
                     auths.put(endpoint.toString(), new JSONObject()
                             .accumulate("auth", getToken())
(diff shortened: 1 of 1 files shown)


shinji62:
@ndeloof any update ?

ndeloof:
:-1: your fix is wrong, it will always use .docker/config.json on new installation, even when docker < 1.7 is in use. This one should only be used when detected, otherwise always prefer the legacy format - or detect docker version to choose the correct one

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryToken.java
```diff
@@ -87,8 +87,8 @@ public Void call() throws IOException {
                         auths = json.getJSONObject("auths");
                     } else {
                         File dockercfg = new File(System.getProperty("user.home"), ".dockercfg");
+                        config = dockercfg;
```
ndeloof: you could fully remove dockercfg variable here

shinji62: Ok will work on that.

shinji62:
Ok, I remove the dockercfg variable.
Should be ok now.

oleg-nenashev:
:+1:

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-29627

oleg-nenashev:
@shinji62 Thanks a lot for your contribution. I've added the plugin to my watch list (BTW I'm not a formal owner).

chmontgomery:
How do I get this fix? Seems just upgrading the plugin didn't work.... do I need to uninstall and reinstall so that it re-creates the `.dockercfg` file?

oleg-nenashev:
@chmontgomery 
You're right, I should have been noticed it during the review.
In order to get the fix, you should re-save all  Docker Registry Credentials. It can be done manually or via Groovy scripts, but probably we need an automatic migration engine for the issue.

CCed @jglick, @carlossg and @rsandell , because their plugins are affected by the issue

shinji62:
Yeah sorry, 
I forget to add this step in the fix, but the .docker/config.json is corrupted so one of the easy solution is to remove it.

Should be in the home directory of your jenkins user.

Thanks.

chmontgomery:
Even deleting my `.docker/config.json` didn't work for me because my login credential form looks different than what you show in your documentation. It's missing the email line:

screen shot 2015-07-31 at 9 14 13 am

And there is no option for a different form: 

screen shot 2015-07-31 at 9 16 44 am

Due to this, the only thing that worked for me was deleting `.docker/config.json` and running `docker login`, which created `.dockercfg`

PR #46, opened 2016-02-05
[FIXED JENKINS-32792] Added check for "sha256:" prefix in image ID

theartoflogic:


jgeiger:
How quick can this get merged?

This stops the plugin from working at all on Docker 1.10 with fingerprinting.

oleg-nenashev:
Looks reasonable, so :+1: for merging it as a temporary solution

oleg-nenashev:
Since this plugin is not being own by me, I'm requesting @reviewbybees

theartoflogic:
> This pull request originates from a CloudBees employee.

I created this pull request and I don't work for CloudBees :)

Regarding the pull request, I'm not entirely happy with the solution as it's not really future-proof as much as it might be able to be. A better solution might be to look for any prefix (i.e. look for a colon in the `id`) and strip it off, since I see in the Docker codebase that introduced this image ID change [a couple of commented-out lines](https://github.com/aaronlehmann/docker/commit/500e77bad0b19b3b1c8e6ac195485adcb70daef1#diff-181c64cf9dc642b144473c58272921f7R21) that hint at other possible prefixes, as well as lengths for the hash after the prefix due to different hashing algorithms being used, such as sha384 and sha512.

If we do want to try to future-proof the solution to account for varying-length hashes then we'd need to change the length check of the `id` as well. I'm still not sure what the purpose of the length check is in this case, so removing it altogether is another potential option.

jgeiger:
My workaround was to remove fingerprinting on all my jenkins Docker builds, but merging this would still help people.

oleg-nenashev:
@theartoflogic 

> I created this pull request and I don't work for CloudBees :)

Yes, that's the downside of the current review bot. We need to fix it later

batmat:
+1 IMO this is an urgent merge + release. 
1.10 has gone GA very recently, and more and more people are gonna be blocked by this as they upgrade.

For example, we just encountered that one using the docker Pipeline DSL.

ping @jenkinsci/code-reviewers

oleg-nenashev:
We have two +1s , so I'm going to merge the PR and kick-off the release

oleg-nenashev:
@theartoflogic 
Thanks for your contribution

batmat:
Thanks OIeg.

2016-02-08 12:49 GMT+00:00 Oleg Nenashev <notifications@github.com>:

oleg-nenashev:
The fix has been released

jglick:
[Background](https://github.com/docker/docker/wiki/Engine-v1.10.0-content-addressability-migration)

PR #52, opened 2016-08-13
[FIXED JENKINS-37390] Added @Symbol to DockerTool.DescriptorImpl

abayer:
[JENKINS-37390](https://issues.jenkins-ci.org/browse/JENKINS-37390)

cc @reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
@@ -43,6 +43,7 @@ import hudson.util.ListBoxModel;
 import jenkins.authentication.tokens.api.AuthenticationTokens;
 import jenkins.model.Jenkins;
 
+import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.AncestorInPath;
 import org.kohsuke.stapler.DataBoundConstructor;
 
@@ -254,7 +255,7 @@ public class DockerRegistryEndpoint extends AbstractDescribableImpl<DockerRegist
         return true;
     }
     
-    @Extension
+    @Extension @Symbol("dockerHubEndpoint")
     public static class DescriptorImpl extends Descriptor<DockerRegistryEndpoint> {
         @Override
         public String getDisplayName() {
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
@@ -28,6 +28,7 @@ import com.cloudbees.plugins.credentials.impl.BaseStandardCredentials;
 import hudson.Extension;
 import hudson.Util;
 import hudson.util.Secret;
+import org.jenkinsci.Symbol;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 import javax.annotation.CheckForNull;
@@ -92,7 +93,7 @@ public class DockerServerCredentials extends BaseStandardCredentials {
         return serverCaCertificate;
     }
 
-    @Extension
+    @Extension @Symbol("dockerHostCert")
     public static class DescriptorImpl extends BaseStandardCredentialsDescriptor {
         @Override
         public String getDisplayName() {
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerDomainSpecification.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerDomainSpecification.java
@@ -29,6 +29,7 @@ import com.cloudbees.plugins.credentials.domains.DomainSpecification;
 import com.cloudbees.plugins.credentials.domains.DomainSpecificationDescriptor;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.Extension;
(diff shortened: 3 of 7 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerTool.java
```diff
@@ -114,7 +115,8 @@ public DockerTool forNode(Node node, TaskListener log) throws IOException, Inter
         }
     }
 
-    @Extension public static class DescriptorImpl extends ToolDescriptor<DockerTool> {
+    @Extension @Symbol("docker")
```
oleg-nenashev: There is also a docker global environment variable in Docker Pipeline plugin. Is it safe?

abayer: Yes - the symbols are scoped by extension point.

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -162,7 +163,7 @@ public Object invoke(File f, VirtualChannel channel) throws IOException, Interru
         return install;
     }
 
-    @Extension
+    @Extension @Symbol("dockerIoInstaller")
```
oleg-nenashev: Uhm, bad naming

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
```diff
@@ -92,7 +93,7 @@ public String getServerCaCertificate() {
         return serverCaCertificate;
     }
 
-    @Extension
+    @Extension @Symbol("dockerHostCert")
```
oleg-nenashev: Maybe dockerServerCertificate() ?

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -254,7 +255,7 @@ public boolean equals(Object obj) {
         return true;
     }
     
-    @Extension
+    @Extension @Symbol("dockerHubEndpoint")
```
jglick: :bug: not necessarily for Hub (in fact usually not), and anyway this class is not `abstract` so the actual symbol should be as short as possible: say, `endpoint`.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerDomainSpecification.java
```diff
@@ -60,7 +61,7 @@ public Result test(DomainRequirement scope) {
     }
 
     /** {@inheritDoc} */
-    @Extension
+    @Extension @Symbol("dockerServerDomain")
```
jglick: Probably do not want to repeat `domain` in the symbol.

jglick: (Though I am not exactly sure what extension point is being implemented here.)

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
```diff
@@ -172,7 +173,7 @@ public boolean equals(Object obj) {
         return true;
     }
     
-    @Extension
+    @Extension @Symbol("dockerDaemonEndpoint")
```
jglick: :bug: ditto, just `endpoint` would suffice

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
abayer: Yeah, I'm gonna let this sit for a day or so for me to think about - I may just yank most of these `@Symbol`s since it's really hard to figure out what they should actually be called...

oleg-nenashev:
So there is 🐛 from @jglick 
@abayer any plans?

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
jglick: > it's really hard to figure out what they should actually be called...

https://github.com/jenkinsci/system-config-dsl-plugin is the only plausible client that I know of for a `@Symbol` on a `ToolInstallerDescriptor`.

oleg-nenashev: This installer does not work in any case :(

jglick: Already fixed in #53.

stephenc:
@abayer ping

michelzanini:
Whats the reason why this PR was closed? 
It would be really useful to have `@Symbol('docker')` on `ToolDescriptor<DockerTool>`.
@abayer

oleg-nenashev:
JCasC does not work well with conflicting symbols as we discovered. So there may be some rework needed with this approach

jglick:
> JCasC does not work well with conflicting symbols as we discovered.

Say what?

PR #54, opened 2016-09-12
credentials-binding implementation for Docker client certificates

thomasgl-orange:
This is an implementation of Binding (the Extension for credentials-binding plugin) for DockerServerCredentials (client certificates). It gives a simple way to make these credentials available in any kind of build step or pipeline jobs, using the usual credentials-binding mecanism. The {ca,cert,key}.pem files will be written in a temporary directory, whose path can be bound to (by default) $DOCKER_CERT_PATH.

This PR depends on a little refactoring in credentials-binding-plugin, to reuse code from their FileBinding implementation - see jenkinsci/credentials-binding-plugin#25. Until it is accepted or rejected, this PR can't be merged, so I'm only submitting it for comments/reference.


oleg-nenashev:
Is it still WiP?

thomasgl-orange:
@oleg-nenashev 

Well, it's still something I would like to see merged in the docker-commons plugin (in the mean time, I'm using a similar implementation in an in-house plugin, but I would happily get rid of it).  But this PR depends on jenkinsci/credentials-binding-plugin#25 being merged, which is still pending... (sure, I could rewrite this without the refactoring to credentials-binding-plugin, but that would mean some code duplicated from FileCredentialsBinding).

One other thing that was left to do was making the JUnit test work on Windows (it was relying on an 'sh' Pipeline step). I've just pushed the necessary changes: there is now an alternative 'bat' step for testing on Windows.

And one last thing which would be nice to add is an @Symbol annotation (to improve the syntax when using Pipeline or Job-DSL), but for that too I'm waiting for changes in credentials-binding-plugin to be merged (jenkinsci/credentials-binding-plugin#24).

oleg-nenashev:
I summon @jglick to the thread

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsBindingTest.java
```diff
@@ -0,0 +1,182 @@
...
+                            + "      set -e -x\n"
+                            + "\n"
+                            + "      # check permissions on the credentials dir and its parent\n"
+                            + "      [ $(stat -c %a \"$DOCKER_CERT_PATH\")    = 700 ]\n"
+                            + "      [ $(stat -c %a \"$DOCKER_CERT_PATH\"/..) = 700 ]\n"
+                            + "\n"
+                            + "      # check permissions and content of the certificate files\n"
+                            + "      [ $(stat -c %a \"$DOCKER_CERT_PATH/key.pem\")  = 600 ]\n"
+                            + "      [ $(stat -c %a \"$DOCKER_CERT_PATH/cert.pem\") = 600 ]\n"
+                            + "      [ $(stat -c %a \"$DOCKER_CERT_PATH/ca.pem\")   = 600 ]\n"
```
jglick: BTW this stuff could be put into `src/test/resources/` and copied prior to running the build using `TheClass.class.getResource[AsStream]` and `Node.getWorkspaceFor`. Might be more readable that way.

pom.xml
```diff
@@ -71,6 +76,73 @@
...
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-basic-steps</artifactId>
+      <version>2.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-scm-step</artifactId>
+      <version>2.3</version>
+      <classifier>tests</classifier>
```
jglick: Deleted in a later version. You can probably delete this dep?

src/main/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsBinding/help-variable.html
```diff
@@ -0,0 +1,5 @@
+<div>
+    Name of an environment variable to be set during the build.<br>
+    Its value will be the absolute path of the directory where the <code>{ca,cert,key}.pem</code> files will be created.<br>
+    You probably want to call this variable <code>DOCKER_CERT_PATH</code>, which will be undestood by the docker client binary.<br>
```
jglick: typo

src/main/resources/org/jenkinsci/plugins/docker/commons/Messages.properties
```diff
@@ -1,2 +1,3 @@
 DockerServerDomainSpecification.DisplayName=Docker Server Credentials
 DockerFingerprintAction.DisplayName=Docker Fingerprints
+DockerServerCredentialsBinding.DisplayName=Docker client certificate
```
jglick: add trailing newline

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsBinding.java
```diff
@@ -0,0 +1,59 @@
...
+import org.kohsuke.stapler.DataBoundConstructor;
+
+import hudson.Extension;
+import hudson.FilePath;
+
+public class DockerServerCredentialsBinding extends AbstractOnDiskBinding<DockerServerCredentials> {
+
+    @DataBoundConstructor
+    public DockerServerCredentialsBinding(String variable, String credentialsId) {
+        super(variable, credentialsId);
```
jglick: Would be possible to extend `MultiBinding` directly (using `UnbindableDir`), which would enable you to supply a default value for `variable` by turning it into a `@DataBoundSetter`. (In the current inheritance chain this is impossible since the field is `final`.) Would enable Pipeline syntax to be simply

```groovy
withCredentials([dockerCert(credentialsId: 'docker-client-cert')]) {
  sh 'docker push'
}
```

Not sure if it is worth the effort to make that refactoring, just FYI.

thomasgl-orange: I like the idea, but, well, I will pass. Lazy me...

I've pushed fixes for your other review comments.

PR #56, opened 2016-10-21
[JENKINS-39181] Invalid fully qualified image name when registry URL specified

liebowitz:
This issue was discovered in the docker-workflow-plugin. 
Given the following pipeline code:

`docker.withRegistry('http://localhost:5000/') {`
`def dockerImage = docker.image('localhost:5000/test-image')`
`dockerImage.inside {`
`echo "container running on ${port}"`
`}`
`}`

The rendered "docker pull" command incorrectly builds the qualified image name as: 
_localhost:5000/localhost:5000/test-image_

This PR adds a condition to check whether or not the image name already contains the registry prefix to avoid appending again.

cc: @genehand

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
@@ -228,8 +228,13 @@ public class DockerRegistryEndpoint extends AbstractDescribableImpl<DockerRegist
      * @return the full registry:port/namespace/name string
      * @throws IOException
      */
-    public String imageName(String userAndRepo) throws IOException {
-        if (url==null)    return userAndRepo;
+    public String imageName(@Nonnull String userAndRepo) throws IOException {
+        if (userAndRepo == null) {
+            throw new IllegalArgumentException("Image name cannot be null.");
+        }
+        if (url == null) {
+            return userAndRepo;
+        }
         URL effectiveUrl = getEffectiveUrl();
 
         StringBuilder s = new StringBuilder(effectiveUrl.getHost());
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -236,6 +236,9 @@ public String imageName(String userAndRepo) throws IOException {
         if (effectiveUrl.getPort() > 0 && effectiveUrl.getDefaultPort() != effectiveUrl.getPort()) {
             s.append(':').append(effectiveUrl.getPort());
         }
+        if (userAndRepo.startsWith(String.valueOf(s))) {
```
oleg-nenashev: Before doing it you need to ensure the field is not null.
There is no related annotations in the code. Maybe makes sense to update method definition

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpointTest.java
```diff
@@ -52,6 +52,12 @@ public void testParseWithTags() throws Exception {
         assertRegistry("https://docker.acme.com", "docker.acme.com/busybox:tag");
     }
 
+    @Test
+    public void testParseFullyQualifiedImageName() throws Exception {
+        assertEquals("private-repo:5000/test-image", new DockerRegistryEndpoint("http://private-repo:5000/", null).imageName("private-repo:5000/test-image"));
+        assertEquals("private-repo:5000/test-image", new DockerRegistryEndpoint("http://private-repo:5000/", null).imageName("test-image"));
+    }
```
oleg-nenashev: add null check if you think Null may really happen. I think it can :(

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
oleg-nenashev: Agreed, better to propagate the error

liebowitz:
@oleg-nenashev Hi Oleg, PR has been open for months. When do you expect to merge? Thanks

oleg-nenashev:
@liebowitz Hi. Sorry, I'm not a maintainer of this repo (though I made a couple of releases). Hence I was not monitoring PRs here. Will merge and release it today

liebowitz:
@oleg-nenashev Thanks for merging. The dependency version on [docker-workflow-plugin](https://github.com/jenkinsci/docker-workflow-plugin/blob/master/pom.xml#L50) will need to be bumped after the release.

oleg-nenashev:
@liebowitz agreed. CC @jglick, who is a maintainer of Docker Pipeline Plugin.

PR #57, opened 2017-01-31
remove check for default port! it should always set the port. This is…

ediri:
… important for private registries with tricky SSL Settings and Reverse Proxy Settings.


oleg-nenashev:
@jenkinsci/code-reviewers

ediri:
no news here?

ediri:
can someone approve this? @oleg-nenashev

oleg-nenashev:
I suppose my vote is enough taking the fact there is no other responses

PR #58, opened 2017-02-16
Fix PCT Test Failures

kwhetstone:
With the new changes to the plugin compatibility tester, a few of this plugin's tests fail, specifically `DockerServerCredentialsTest` and `DockerServerDomainSpecificationTest`. Currently, it depends on the 1.x line of the credentials plugin. Since there were significant changes with this plugin upon the 2.x release & the plugin compat tester tests compatibility against the most recent or bundled version of the plugin, the tests are failing even though there are really no issues. I have run all the docker-commons tests using `mvn test` as well as through the plugin compat tester, and there are now no errors.

Removed the check of a url that is incompatable across a major release.  This type of check can be done in the acceptance test harness.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,11 +56,6 @@
       <version>2.1.11</version>
     </dependency>
     <dependency>
-      <groupId>args4j</groupId>
-      <artifactId>args4j</artifactId>
-      <version>2.0.31</version>
-    </dependency>
-    <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>authentication-tokens</artifactId>
       <version>1.1</version>
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.580.3</jenkins.version>
+    <jenkins.version>1.590</jenkins.version>
     <java.level>6</java.level>
   </properties>
 
@@ -54,6 +54,11 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials</artifactId>
       <version>2.1.11</version>
+    </dependency>
+    <dependency>
+      <groupId>args4j</groupId>
+      <artifactId>args4j</artifactId>
+      <version>2.0.31</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -31,7 +31,7 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.590</jenkins.version>
+    <jenkins.version>1.609.3</jenkins.version>
     <java.level>6</java.level>
   </properties>
 
--- a/pom.xml
+++ b/pom.xml
@@ -53,7 +53,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials</artifactId>
-      <version>1.22</version>
+      <version>2.1.11</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


kwhetstone:
I knew this was a big change.  The other tests in the suite pass with the credentials upgrade, so this is just a case of changing the checks.

oleg-nenashev:
We also need to ensure that Credentials API does not cause a fallout in other Docker plugins depending on this lib.

kwhetstone:
There looks like there's an args4j dependency issue here (fixing), but I don't know how this plugin works with the other docker plugins.  Is this a check I can make with some acceptance tests?  Is this going to lead to all the plugins needing an upgrade to the latest credentials?

kwhetstone:
I had to bump the dependency to 1.590 to pull in getActiveInstance.  I could move to 1.609, but I didn't want to move the core version too much.  From what I can tell this won't affect the other docker plugins since they don't appear to all be dependent on each other.

pom.xml
```diff
@@ -31,7 +31,7 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.580.3</jenkins.version>
+    <jenkins.version>1.590</jenkins.version>
```
jglick: Just go to the next LTS line, 1.609.3 (or later, if important downstream projects are already on something newer anyway).

kwhetstone:
Bumped to 1.609.3.

kwhetstone:
Hey everyone, is there anything we can do to move this forward?

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsTest.java
```diff
@@ -57,7 +57,8 @@ public void configRoundTripEmpty() throws Exception {
                 Collections.<DomainSpecification>singletonList(new DockerServerDomainSpecification()));
         DockerServerCredentials credentials = new DockerServerCredentials(CredentialsScope.GLOBAL, "foo", "desc", "", "", "");
         store.addDomain(domain, credentials);
-        j.submit(j.createWebClient().goTo("credential-store/domain/" + domain.getName() + "/credential/"+credentials.getId()+"/update")
+
+        j.submit(j.createWebClient().goTo("credentials/store/system/domain/" + domain.getName() + "/credential/"+credentials.getId()+"/update")
```
oleg-nenashev: I am surprised you have to edit the links here. Have the original links been deprecated in Credentials2.x ?

kwhetstone: Yes, it just ends up throwing a 404.  I even checked on running copy of Jenkins too in order to see if there was a redirect that wasn't followed.  Nothing.

oleg-nenashev: @stephenc is it an intended compat change?

kwhetstone:
Hey everyone, anything else on this PR?  I think the credentials 404 is the only question left, however that's outside this plugin.  Is there anything else I should address here?

oleg-nenashev:
I suppose we need a response from @stephenc , and then the PR is fine

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsTest.java
jglick: I do not think these were a documented REST API to begin with.

jglick:
Nothing wrong with this PR, so I am merging. If someone thinks there is a compatibility bug in `credentials`, file it.

PR #59, opened 2017-08-21
[JENKINS-46321] Avoid 'JSONException: null object' when no key 'auths' exist in .docker/config.json

shyr:
To avoid 'JSONException: null object' error, if there is no key 'auths' in the `config.json` file, this PR adds that key and writes it to the file in advance.

Even though there is no 'auths' key at `config.json` file, docker client adds it when it is needed.
This plugin should be able to avoid that situation, too.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryToken.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryToken.java
@@ -76,12 +76,12 @@ public final class DockerRegistryToken implements Serializable {
      * Makes the credentials available locally and returns {@link KeyMaterialFactory} that gives you the parameters
      * needed to access it.
      *
-     * This is done by inserting the token into {@code ~/.dockercfg}
+     * This is done by inserting the token into {@code ~/.dockercfg} or {@code ~/.docker/config.json}
      */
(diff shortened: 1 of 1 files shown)


shyr:
Please take a look.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryToken.java
```diff
@@ -96,6 +96,11 @@ public Void call() throws IOException {
                     if (config.exists()) {
                         json = JSONObject.fromObject(FileUtils.readFileToString(config, "UTF-8"));
                         auths = json.getJSONObject("auths");
+                        if (auths.isNullObject()) {
+                            auths = new JSONObject();
+                            json.put("auths", auths);
+                            FileUtils.writeStringToFile(config, json.toString(2), "UTF-8");
```
oleg-nenashev: I would add some logic on the FINE/CONFIG level here. It is just a data migration logic IIUC

jglick:
Might be better than nothing (no idea, really) but [JENKINS-46321](https://issues.jenkins-ci.org/browse/JENKINS-46321) would just disappear if [JENKINS-38018](https://issues.jenkins-ci.org/browse/JENKINS-38018) were done instead.

jglick:
Likely unnecessary after #67, though could still apply to corner cases such as users of the Google container registry which cannot be handled by a simple `docker login`.

PR #61, opened 2017-10-10
Remove icon-shim dependency

jglick:
Not sure why it was here—was actually older than what you would get by default anyway. Cf. https://github.com/jenkinsci/plugin-pom/pull/72.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.15</version>
+    <version>2.36</version>
     <relativePath />
   </parent>
   <groupId>org.jenkins-ci.plugins</groupId>
@@ -19,7 +19,7 @@
   <licenses>
     <license>
       <name>MIT License</name>
-      <url>http://opensource.org/licenses/MIT</url>
+      <url>https://opensource.org/licenses/MIT</url>
     </license>
   </licenses>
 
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -33,6 +33,7 @@
   <properties>
     <jenkins.version>1.642.4</jenkins.version>
     <java.level>7</java.level>
+    <jenkins-test-harness.version>2.30-20171010.153224-1</jenkins-test-harness.version> <!-- TODO https://github.com/jenkinsci/jenkins-test-harness/pull/78 -->
```
oleg-nenashev: I would propose to release JTH and Plugin POM before merging this PR

jglick: yes, would do

jglick: Well, JTH at least. `plugin-pom` can come along later, not urgent.

KostyaSha:
> Not sure why it was here

Because of docker-traceability. Lol, seems I remember introduced docker chaos better.

KostyaSha:
So now you broke dependent plugins??
I guess identity was needed for  https://github.com/jenkinsci/docker-traceability-plugin/blob/master/docker-traceability-plugin/pom.xml#L141 
And icon shim for docker icons.

jglick:
It should still get `icon-shim` from core. A slightly later version in fact.

KostyaSha:
@jglick Then it looks safe.  Afair this plugin was created as piece of traceability and later reused in other plugins.

PR #62, opened 2017-12-08
Create directory before calling chmod

dwnusbaum:
See [JENKINS-36088](https://issues.jenkins-ci.org/browse/JENKINS-36088?focusedCommentId=322239&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-322239) and subsequent comments. Before Jenkins 2.93, calling `FilePath#chmod` on a non-existent file or directory would fail silently, now it will throw `NoSuchFileException`. There appears to be an incorrect usage here that went unnoticed because the calls to`ServerKeyMaterialFactory#copyInto` call `Files#write` which creates parent directories if they do not exist.

I'm not sure why `DockerServerCredentialsBindingTest#basics` passes without this change. It runs [basics-step1.sh](https://github.com/jenkinsci/docker-commons-plugin/blob/master/src/test/resources/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsBindingTest/basics-step1.sh) which checks folder and file permissions in a build and asserts that the build succeeds. The test passes locally for me no matter what permissions are passed to `FilePath#chmod` in `ServerKeyMaterialFactory`.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/impl/ServerKeyMaterialFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/impl/ServerKeyMaterialFactory.java
@@ -81,9 +81,10 @@ public class ServerKeyMaterialFactory extends KeyMaterialFactory {
 
         if (key != null && cert != null && ca != null) {
             final FilePath tempCredsDir = new FilePath(getContext().getBaseDir(), UUID.randomUUID().toString());
+            tempCredsDir.mkdirs();
 
             // protect this information from prying eyes
-            tempCredsDir.chmod(0600);
+            tempCredsDir.chmod(0700);
 
             // these file names are defined by convention by docker
             copyInto(tempCredsDir, "key.pem", key);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/impl/ServerKeyMaterialFactory.java
```diff
@@ -81,9 +81,10 @@ public KeyMaterial materialize() throws IOException, InterruptedException {
 
         if (key != null && cert != null && ca != null) {
             final FilePath tempCredsDir = new FilePath(getContext().getBaseDir(), UUID.randomUUID().toString());
+            tempCredsDir.mkdirs();
 
             // protect this information from prying eyes
-            tempCredsDir.chmod(0600);
+            tempCredsDir.chmod(0700);
```
dwnusbaum: Now that the chmod call succeeds, using `0600` causes `DockerServerEndpointTest` to fail because the files in the folder are inaccessible if the executable bit is not set.

jglick:
This was filed as [JENKINS-48453](https://issues.jenkins-ci.org/browse/JENKINS-48453) BTW.

PR #63, opened 2018-01-04
JENKINS-48674: downloads image for newer version from download.docker.com

denisa:
Update the plugin to use the new host and url scheme while preserving access to older images (pre 17.05.0-ce).

https://issues.jenkins-ci.org/browse/JENKINS-48674

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
@@ -34,10 +34,14 @@ import hudson.remoting.VirtualChannel;
 import hudson.tools.ToolInstallation;
 import hudson.tools.ToolInstaller;
 import hudson.tools.ToolInstallerDescriptor;
+import hudson.util.VersionNumber;
 import java.io.IOException;
 import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
 import java.net.URL;
 import java.net.URLConnection;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import javax.annotation.Nonnull;
 import jenkins.security.MasterToSlaveCallable;
 import org.kohsuke.stapler.DataBoundConstructor;
@@ -68,7 +72,7 @@ public class DockerToolInstaller extends ToolInstaller {
         }
         String os = nodeChannel.call(new FindArch());
 
-        final URL url = new URL("https://get.docker.com/builds/"+os+"/docker-"+version);
+        final URL url = getDockerImageUrl(os, version);
         FilePath install = preferredLocation(tool, node);
 
         // (simplified) copy/paste from FilePath as hudson.FilePath.installIfNecessaryFrom do assume the URL points to be a zip/tar archive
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -142,6 +146,27 @@ public FilePath performInstallation(ToolInstallation toolInstallation, @Nonnull
...
+        String osName="";
+        if (os.startsWith("linux")) osName = "Linux";
+        if (os.startsWith("win")) osName = "Windows";
+        if (os.startsWith("mac")) osName = "Darwin";
+        return new URL("https://get.docker.com/builds/" + osName + os.substring(os.lastIndexOf("/")) + "/docker-" + version);
+    }
+
+    private static VersionNumber parseVersion(String version) {
+        // any version that sorts before 17.05.0-ce
+        if (version.equals("latest")) return new VersionNumber("0");
```
oleg-nenashev: I would rather prefer an API with `null` and annotations, but fine for a private method

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -142,6 +146,27 @@ public FilePath performInstallation(ToolInstallation toolInstallation, @Nonnull
...
     }
 
+    static URL getDockerImageUrl(String os, String version) throws MalformedURLException {
+        if (parseVersion(version).isNewerThan(parseVersion("17.05.0-ce")))
+            return new URL("https://download.docker.com/" + os + "/docker-" + version);
+
+        String osName="";
+        if (os.startsWith("linux")) osName = "Linux";
+        if (os.startsWith("win")) osName = "Windows";
+        if (os.startsWith("mac")) osName = "Darwin";
```
oleg-nenashev: Would it make sense to print warnings if there is no match? (e.g. new platform)

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
```diff
@@ -142,6 +146,27 @@ public FilePath performInstallation(ToolInstallation toolInstallation, @Nonnull
         return install;
     }
 
+    static URL getDockerImageUrl(String os, String version) throws MalformedURLException {
+        if (parseVersion(version).isNewerThan(parseVersion("17.05.0-ce")))
```
oleg-nenashev: `isOlderThan()`?

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstaller.java
jglick: Would `latest` not be “newer than” `17.05.0-ce`? Your test does request this go to `get.docker.com`. Yet https://get.docker.com/builds/Linux/x86_64/docker-latest is a 403. OTOH https://download.docker.com/linux/static/stable/x86_64/docker-latest is a 404. It seems `latest` just does not work any more—the index of https://download.docker.com/linux/static/ lists `edge`, `stable`, and `test`, but the file names in each specify a version. However I will check the test behavior and see what is going on.

jglick: Ah, https://get.docker.com/builds/Linux/x86_64/docker-latest.tgz is a 200.

jglick: …and it is `17.05.0-ce`. So they are still publishing to `get.docker.com`, just not the actual “latest” (which seems to be `17.12.0-ce` currently). Weird.

src/test/java/org/jenkinsci/plugins/docker/commons/tools/DockerToolInstallerTest.java
```diff
@@ -56,35 +74,40 @@ public void smokes() throws Exception {
...
+        try {
+            new URL("https://download.docker.com/").openStream().close();
+        } catch (IOException x) {
+            Assume.assumeNoException("Cannot contact download.docker.com, perhaps test machine is not online", x);
+        }
         r.jenkins.getDescriptorByType(DockerTool.DescriptorImpl.class).setInstallations(
             new DockerTool("latest", "", Collections.singletonList(new InstallSourceProperty(Collections.singletonList(new DockerToolInstaller("", "latest"))))),
-            new DockerTool("1.10.0", "", Collections.singletonList(new InstallSourceProperty(Collections.singletonList(new DockerToolInstaller("", "1.10.0"))))));
+            new DockerTool("1.10.0", "", Collections.singletonList(new InstallSourceProperty(Collections.singletonList(new DockerToolInstaller("", "1.10.0"))))),
+            new DockerTool("17.09.1-ce", "", Collections.singletonList(new InstallSourceProperty(Collections.singletonList(new DockerToolInstaller("", "17.09.1-ce.tgz"))))));
```
jglick: The `version` here is incorrect.

jglick:
I am working on a fixed PR.

PR #64, opened 2018-01-04
[JENKINS-48674] Fix installer

jglick:
Amends #63 by @denisa.

@reviewbybees


denisa:
Thanks for the improvements!
I did too changes to #63 this morning:
(1) gets the binaries from the edge instead of the stable channel.
The edge channel is a superset of the stable channel with more frequent (less tested?) releases.
This gives the user more versions to select from

(2) refactored the code so that FindArch is only concerned with architecture, and getDockerImageUrl concerned with the structure of the URL.

It would be great if change (1), at least, were included.

jglick:
Merged in your latest work and will release the aggregate assuming tests pass. Thanks!

PR #66, opened 2018-01-22
[JENKINS-49075] Bump parent pom and a few other dependencies

rysteboe:
@reviewbybees 

@jglick, bumping the required core version just to satisfy the PCT doesn't sit well with me, but filing the PR anyway for you to accept or reject.

https://issues.jenkins-ci.org/browse/JENKINS-49075

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>2.36</version>
+    <version>3.2</version>
     <relativePath />
   </parent>
   <groupId>org.jenkins-ci.plugins</groupId>
@@ -31,9 +31,8 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.642.4</jenkins.version>
-    <java.level>7</java.level>
-    <jenkins-test-harness.version>2.30</jenkins-test-harness.version>
+    <jenkins.version>2.60.3</jenkins.version>
+    <java.level>8</java.level>
   </properties>
 
   <repositories>
@@ -54,7 +53,7 @@
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>credentials</artifactId>
-      <version>2.1.11</version>
+      <version>2.1.13</version>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins</groupId>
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -79,8 +80,9 @@ private DockerFingerprints() {} // no instantiation
      * @return Created fingerprint or null if it is not found
      * @throws IOException Fingerprint loading error
      */
+    @SuppressFBWarnings("RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE")
     public static @CheckForNull Fingerprint of(@Nonnull String id) throws IOException {
-        final Jenkins jenkins = Jenkins.getInstance(); // should be not null
+        final Jenkins jenkins = Jenkins.getInstance(); // should be not null, but technically can be
```
oleg-nenashev: Use `getInstanceOrNull` then

pom.xml
```diff
@@ -31,9 +31,8 @@
   </scm>
 
   <properties>
-    <jenkins.version>1.642.4</jenkins.version>
-    <java.level>7</java.level>
-    <jenkins-test-harness.version>2.30</jenkins-test-harness.version>
+    <jenkins.version>2.60.3</jenkins.version>
```
jglick: Any particular reason you are pushing it this high? I have no problem with it, but you seemed to be saying that you felt compelled to do so, which I am not following.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: Right, or drop the null check—I see no reason for this code to be run after Jenkins shutdown so the null check should not be necessary.

pom.xml
rysteboe: The PCT fails without this bump.  I can do more investigation into alternative fixes, but this was the only way that I could fix it without spending an undue amount of time working on this.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -79,9 +80,8 @@ private DockerFingerprints() {} // no instantiation
      * @return Created fingerprint or null if it is not found
      * @throws IOException Fingerprint loading error
      */
-    public static @CheckForNull Fingerprint of(@Nonnull String id) throws IOException {
-        final Jenkins jenkins = Jenkins.getInstance(); // should be not null
-        return jenkins != null ? jenkins.getFingerprintMap().get(getFingerprintHash(id)) : null;
+    public static Fingerprint of(@Nonnull String id) throws IOException {
```
jglick: You need to leave `@CheckForNull` on this method—there may be no such fingerprint. (I think you got confused by removing the code path where `Jenkins.instance == null`.)

rysteboe: Yeah fixed.  Didn't look closely enough at the rest of the code.

PR #68, opened 2018-03-28
[JENKINS-48437] Authorization for Docker Endpoint

Dohbedoh:
[JENKINS-48437](https://issues.jenkins-ci.org/browse/JENKINS-48437)

Proposed a fix for the issue with the method that retrieves the credentials to get the DockerRegistryToken.
Add implementation accepting a `Run` as context (and using [CredentialsProvider.findCredentialById](https://github.com/jenkinsci/credentials-plugin/blob/master/src/main/java/com/cloudbees/plugins/credentials/CredentialsProvider.java#L855-L871)) that can be later leverage by docker-workflow.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
@@ -38,6 +38,8 @@ import hudson.model.AbstractDescribableImpl;
 import hudson.model.Describable;
 import hudson.model.Descriptor;
 import hudson.model.Item;
+import hudson.model.Run;
+import hudson.model.TaskListener;
 import hudson.remoting.VirtualChannel;
 import hudson.util.ListBoxModel;
 import jenkins.authentication.tokens.api.AuthenticationTokens;
@@ -62,7 +64,6 @@ import static com.cloudbees.plugins.credentials.CredentialsMatchers.*;
 import hudson.AbortException;
 import hudson.EnvVars;
 import hudson.Launcher;
-import hudson.model.TaskListener;
 import hudson.slaves.WorkspaceList;
 import org.jenkinsci.plugins.docker.commons.tools.DockerTool;
 
@@ -188,6 +189,31 @@ public class DockerRegistryEndpoint extends AbstractDescribableImpl<DockerRegist
     }
 
     /**
+     * Plugins that want to refer to a {@link IdCredentials} should do so via ID string,
+     * and use this method to resolve it and convert to {@link DockerRegistryToken}.
+     *
+     * @param context
+     *       If you are a build step trying to access DockerHub in the context of a build/job,
+     *       specify that build. Otherwise null. If you are scoped to something else, you might
+     *       have to interact with {@link CredentialsProvider} directly.
+     */
+    public @CheckForNull DockerRegistryToken getToken(Run context) {
+        if (credentialsId == null) {
+            return null;
+        }
+
+        List<DomainRequirement> requirements = Collections.emptyList();
+        try {
+            requirements = Collections.<DomainRequirement>singletonList(new HostnameRequirement(getEffectiveUrl().getHost()));
+        } catch (IOException e) {
+            // shrug off this error and move on. We are matching with ID anyway.
+        }
+
+        return AuthenticationTokens.convert(DockerRegistryToken.class,
+                CredentialsProvider.findCredentialById(credentialsId, IdCredentials.class, context, requirements));
+    }
+
+    /**
      * @deprecated Call {@link #newKeyMaterialFactory(Item, FilePath, Launcher, EnvVars, TaskListener, String)}
      */
     @Deprecated
@@ -245,6 +271,24 @@ public class DockerRegistryEndpoint extends AbstractDescribableImpl<DockerRegist
             throw new AbortException("Could not find credentials matching " + credentialsId);
         }
         return token.newKeyMaterialFactory(getEffectiveUrl(), workspace, launcher, env, listener, dockerExecutable);
+    }
+
+    /**
+     * Makes the credentials available locally and returns {@link KeyMaterialFactory} that gives you the parameters
+     * needed to access it.
+     * @param context The build trying to access DockerHub
+     * @param workspace a workspace being used for operations ({@link WorkspaceList#tempDir} will be applied)
+     * @param dockerExecutable as in {@link DockerTool#getExecutable}, with a 1.8+ client
+     */
+    public KeyMaterialFactory newKeyMaterialFactory(@CheckForNull Run context, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull TaskListener listener, @Nonnull String dockerExecutable) throws IOException, InterruptedException {
+        if (credentialsId == null) {
+            return KeyMaterialFactory.NULL; // nothing needed to be done
+        }
+        DockerRegistryToken token = getToken(context);
+        if (token == null) {
+            throw new AbortException("Could not find credentials matching " + credentialsId);
+        }
+        return token.newKeyMaterialFactory(getEffectiveUrl(), workspace, launcher, listener, dockerExecutable);
     }
 
     /**
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
@@ -43,6 +43,7 @@ import hudson.model.Run;
 import hudson.model.TaskListener;
 import hudson.model.queue.Tasks;
 import hudson.remoting.VirtualChannel;
+import hudson.security.ACL;
 import hudson.util.ListBoxModel;
 import jenkins.authentication.tokens.api.AuthenticationTokens;
 import jenkins.model.Jenkins;
@@ -58,6 +59,7 @@ import javax.annotation.Nullable;
 import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.regex.Matcher;
@@ -164,6 +166,9 @@ public class DockerRegistryEndpoint extends AbstractDescribableImpl<DockerRegist
      * Plugins that want to refer to a {@link IdCredentials} should do so via ID string,
      * and use this method to resolve it and convert to {@link DockerRegistryToken}.
      *
+     * Implements the logic {@link CredentialsProvider#findCredentialById(String, Class, Run, DomainRequirement...)}
+     * but for an {@link Item}.
+     *
      * @param context
      *       If you are a build step trying to access DockerHub in the context of a build/job,
      *       specify that job. Otherwise null. If you are scoped to something else, you might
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -182,10 +185,35 @@ DockerRegistryToken getToken(Item context) {
 
         // look for subtypes that know how to create a token, such as Google Container Registry
         return AuthenticationTokens.convert(DockerRegistryToken.class, firstOrNull(CredentialsProvider.lookupCredentials(
-                IdCredentials.class, context, Jenkins.getAuthentication(),requirements),
+                IdCredentials.class, context, Tasks.getAuthenticationOf((Queue.Task)context),requirements),
```
jglick: Smells wrong. As you can see from the Javadoc of that method, it is a _prediction_, not an indication of what _actual_ authentication is associated with the build—which would be `Jenkins.getAuthentication()`.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
```diff
@@ -129,6 +125,26 @@ public KeyMaterialFactory newKeyMaterialFactory(@Nonnull Item context, @Nonnull
...
+    public KeyMaterialFactory newKeyMaterialFactory(@Nonnull Run context, @Nonnull VirtualChannel target) throws IOException, InterruptedException {
+        DockerServerCredentials creds=null;
+        if (credentialsId!=null) {
+            List<DomainRequirement> domainRequirements = URIRequirementBuilder.fromUri(getUri()).build();
+            domainRequirements.add(new DockerServerDomainRequirement());
+            creds = CredentialsProvider.findCredentialById(credentialsId, DockerServerCredentials.class, context,
+                    domainRequirements);
+        }
+
+        // the directory needs to be outside workspace to avoid prying eyes
```
jglick: If a new overload is indeed necessary, clearly we would want the duplicated code to be factored out into an internal method somehow.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
Dohbedoh: This is an attempt to fix the JENKINS-48437 without needing to update the dependencies in `docker-worflow` (or other dependents of `docker-commons`).

Dohbedoh:
@jglick Are the PR builds deploying snapshots that I can use ? How exactly can I do the following ?:

> there is no downstream PR in docker-workflow using a timestamped snapshot to consume the newly introduced APIs

jglick:
> Are the PR builds deploying snapshots that I can use?

No; traditionally this was done by a manual `mvn deploy`. Now that JEP-305 is live, though, I could just “[incrementalify](https://github.com/jenkinsci/incrementals-tools/blob/master/README.md)” this repository.

jglick:
…#70

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
Dohbedoh: @jglick How do you want me to factor this out exactly ? I don't see how I could avoid to duplicate the code here unless I use a typing like `Object` or one of the interface that `Run` and `Item` have in common (not sure if doing this would make sense though). So is it worth it ?

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -247,6 +288,24 @@ public KeyMaterialFactory newKeyMaterialFactory(@CheckForNull Item context, @Non
...
     }
 
+    /**
+     * Makes the credentials available locally and returns {@link KeyMaterialFactory} that gives you the parameters
+     * needed to access it.
+     * @param context The build trying to access DockerHub
+     * @param workspace a workspace being used for operations ({@link WorkspaceList#tempDir} will be applied)
+     * @param dockerExecutable as in {@link DockerTool#getExecutable}, with a 1.8+ client
+     */
+    public KeyMaterialFactory newKeyMaterialFactory(@CheckForNull Run context, @Nonnull FilePath workspace, @Nonnull Launcher launcher, @Nonnull EnvVars env, @Nonnull TaskListener listener, @Nonnull String dockerExecutable) throws IOException, InterruptedException {
```
jglick: Please deprecate the old overload so we can keep track of which one is actually called from `docker-workflow`.

(I have a mind to just move those two steps here, actually, so `docker-workflow` can be left to rot.)

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
jglick: I guess it does not matter so long as only one nondeprecated code path remains.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -181,12 +190,44 @@ DockerRegistryToken getToken(Item context) {
...
+    /**
+     * Plugins that want to refer to a {@link IdCredentials} should do so via ID string,
+     * and use this method to resolve it and convert to {@link DockerRegistryToken}.
+     *
+     * @param context
+     *       If you are a build step trying to access DockerHub in the context of a build/job,
+     *       specify that build. Otherwise null. If you are scoped to something else, you might
+     *       have to interact with {@link CredentialsProvider} directly.
+     */
+    public @CheckForNull DockerRegistryToken getToken(Run context) {
```
jglick: Why is this `public` if it is only called internally and from tests? Make it default access. If and when some other plugin actually needs to call this (apparently `docker-workflow` does not), a separate set of PRs can be made which demonstrates the usage.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -181,12 +190,44 @@ DockerRegistryToken getToken(Item context) {
...
                 allOf(AuthenticationTokens.matcher(DockerRegistryToken.class), withId(credentialsId))));
     }
 
+    /**
+     * Plugins that want to refer to a {@link IdCredentials} should do so via ID string,
+     * and use this method to resolve it and convert to {@link DockerRegistryToken}.
+     *
+     * @param context
+     *       If you are a build step trying to access DockerHub in the context of a build/job,
+     *       specify that build. Otherwise null. If you are scoped to something else, you might
```
jglick: > Otherwise null.

Then mark it `@CheckForNull`.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -181,12 +190,44 @@ DockerRegistryToken getToken(Item context) {
             // shrug off this error and move on. We are matching with ID anyway.
         }
 
+        Authentication runAuth = context instanceof Queue.Task ?
```
jglick: Why it `getToken(Item)` being patched if this method no longer being used?

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -181,12 +190,44 @@ DockerRegistryToken getToken(Item context) {
...
+    public @CheckForNull DockerRegistryToken getToken(Run context) {
+        if (credentialsId == null) {
+            return null;
+        }
+
+        List<DomainRequirement> requirements = Collections.emptyList();
+        try {
+            requirements = Collections.<DomainRequirement>singletonList(new HostnameRequirement(getEffectiveUrl().getHost()));
+        } catch (IOException e) {
+            // shrug off this error and move on. We are matching with ID anyway.
```
jglick: still best to log a warning

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpointTest.java
```diff
@@ -27,14 +27,31 @@
...
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.MockAuthorizationStrategy;
 
 /**
  * @author Carlos Sanchez <carlos@apache.org>
  */
 public class DockerRegistryEndpointTest {
 
+    @Rule
+    public JenkinsRule j = new JenkinsRule();
```
jglick: In this case please mark existing test cases `@WithoutJenkins`.

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpointTest.java
```diff
@@ -72,6 +89,56 @@ public void testParseNullUrlAndImageName() throws Exception {
         new DockerRegistryEndpoint(null, null).imageName(null);
     }
 
+    @Issue("JENKINS-48437")
+    @Test
+    public void testGetTokenForItem() throws IOException {
```
jglick: Do we care about this case? IIUC it is deprecated anyway.

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpointTest.java
```diff
@@ -72,6 +89,56 @@ public void testParseNullUrlAndImageName() throws Exception {
...
+                .grant(Computer.BUILD).everywhere().to("alice")
+                .grant(Item.CONFIGURE).everywhere().to("alice");
+        j.jenkins.setAuthorizationStrategy(auth);
+
+        String globalCredentialsId = "global-creds";
+        IdCredentials credentials = new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL,
+                globalCredentialsId, "test-global-creds", "user", "password");
+        CredentialsProvider.lookupStores(j.jenkins).iterator().next().addCredentials(Domain.global(), credentials);
+
+        FreeStyleProject item = j.createFreeStyleProject("testGetToken");
```
jglick: BTW `@Rule JenkinsRule` starts a new instance for each test case, so there is no need to pass a job name argument.

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpointTest.java
```diff
@@ -72,6 +89,56 @@ public void testParseNullUrlAndImageName() throws Exception {
...
+
+        SecurityContextHolder.getContext().setAuthentication(User.get("alice").impersonate());
```
jglick: Please rather use `ACL.as`, even in tests. Just safer.

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpointTest.java
```diff
@@ -27,14 +27,31 @@
...
 
+import com.cloudbees.plugins.credentials.CredentialsProvider;
+import com.cloudbees.plugins.credentials.CredentialsScope;
+import com.cloudbees.plugins.credentials.common.IdCredentials;
+import com.cloudbees.plugins.credentials.domains.Domain;
+import com.cloudbees.plugins.credentials.impl.UsernamePasswordCredentialsImpl;
+import hudson.model.*;
+import jenkins.model.Jenkins;
+import org.acegisecurity.context.SecurityContextHolder;
+import org.apache.commons.codec.binary.Base64;
```
jglick: BTW there is a standard utility now in Java 8.

jglick:
> there is no functional test. (I never managed to write one in `DockerRegistryEndpointTest`

`RegistryEndpointStepTest` I meant.

…though you _could_ write a downstream functional test demonstrating the complete scenario in `ServerEndpointStepTest`, using `variables` as a model (no need for the Jenkins restart in the middle), since in that case the step does more portable stuff which you can test comfortably using `MockQueueItemAuthenticator` and a user credentials store.

mfominov:
@jglick any plans to accept this PR or fix this bug by Jenkins core team?

Dohbedoh:
Adding tests downstream: https://github.com/jenkinsci/docker-workflow-plugin/pull/141

oleg-nenashev:
Jenkins core team has nothing to do with this plugin. Individual
contributors might be interested, but AFAIK there is no active maintainer
here

caitlin615:
Hoping to get an update on releasing this? It's currently blocking me from using docker within pipelines with authentication. Thanks in advance!

PR #69, opened 2018-04-03
[JENKINS-41880] Synchronise on Fingerprint

bjoernhaeuser:
I am not sure this makes any sense. I am just trying to solve this
issue: https://issues.jenkins-ci.org/browse/JENKINS-41880

My thoughts:

This is likely caused by changing the finger prints, the finger prints
are scoped to the project (right?). If this is the case synchronising on
the project could make sense. Also `addFromFacet` should not be a
performance killer and rather quick.

The alternative would be for me to synchronise the complete access to the method.

WDYT?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
@@ -96,20 +96,20 @@ public class DockerFingerprintAction implements RunAction2, IconSpec {
     public @CheckForNull String getFingerprintHash(@CheckForNull String imageId) {
         return (imageId != null) ? DockerFingerprints.getFingerprintHash(imageId) : null;
     }
-    
+
     @Restricted(NoExternalUse.class)
     public @CheckForNull Fingerprint getFingerprint(@CheckForNull String imageId) {
         if (imageId == null) {
             return null;
         }
-        
+
         try {
             return DockerFingerprints.of(imageId);
         } catch (IOException ex) {
             return null; // nothing to do in web UI - return null as well
         }
     }
-    
+
     public List<DockerFingerprintFacet> getDockerFacets(String imageId) {
         List<DockerFingerprintFacet> res = new LinkedList<DockerFingerprintFacet>();
         final Fingerprint fp = getFingerprint(imageId);
@@ -125,9 +125,9 @@ public class DockerFingerprintAction implements RunAction2, IconSpec {
 
     /**
      * Adds an action with a reference to fingerprint if required. It's
-     * recommended to call the method from {
+     * recommended to call the method from {@link hudson.BulkChange}
+     * transaction to avoid saving the {@link Run} multiple times.
      *
-     * @BulkChange} transaction to avoid saving the {@link Run} multiple times.
      * @param fp Fingerprint
      * @param imageId ID of the docker image
      * @param run Run to be updated
@@ -140,7 +140,7 @@ public class DockerFingerprintAction implements RunAction2, IconSpec {
                 action = new DockerFingerprintAction();
                 run.addAction(action);
             }
-            
+
             if (action.imageIDs.add(imageId)) {
                 run.save();
             } // else no need to save updates
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
@@ -96,20 +96,20 @@ public class DockerFingerprintAction implements RunAction2, IconSpec {
     public @CheckForNull String getFingerprintHash(@CheckForNull String imageId) {
         return (imageId != null) ? DockerFingerprints.getFingerprintHash(imageId) : null;
     }
-
+    
     @Restricted(NoExternalUse.class)
     public @CheckForNull Fingerprint getFingerprint(@CheckForNull String imageId) {
         if (imageId == null) {
             return null;
         }
-
+        
         try {
             return DockerFingerprints.of(imageId);
         } catch (IOException ex) {
             return null; // nothing to do in web UI - return null as well
         }
     }
-
+    
     public List<DockerFingerprintFacet> getDockerFacets(String imageId) {
         List<DockerFingerprintFacet> res = new LinkedList<DockerFingerprintFacet>();
         final Fingerprint fp = getFingerprint(imageId);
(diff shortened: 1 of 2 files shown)


bjoernhaeuser:
Thanks!

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprintAction.java
```diff
@@ -96,20 +96,20 @@ public String getUrlName() {
     public @CheckForNull String getFingerprintHash(@CheckForNull String imageId) {
         return (imageId != null) ? DockerFingerprints.getFingerprintHash(imageId) : null;
     }
-    
+
```
jglick: Please revert formatting-only changes, and configure your editor to not change whitespace in lines you have not otherwise edited.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -254,55 +254,57 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
...
-        for (FingerprintFacet facet : facets) {
-            if (facet instanceof DockerAncestorFingerprintFacet) {
-                ancestorFacet = (DockerAncestorFingerprintFacet) facet;
-                break;
-            }
-        }
-        BulkChange bc = new BulkChange(f);
-        try {
-            if (ancestorFacet == null) {
-                ancestorFacet = new DockerAncestorFingerprintFacet(f, timestamp, descendantImageId);
```
jglick: Unreadable diff. [This](https://github.com/jenkinsci/docker-commons-plugin/pull/69/files?w=1) is much easier to follow, but generally I recommend not changing indentation of large code blocks, since it messes up other things like cherry-picks and merges.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -26,19 +26,19 @@
 import hudson.BulkChange;
 import hudson.model.Fingerprint;
 import hudson.model.Run;
+import jenkins.model.FingerprintFacet;
 import jenkins.model.Jenkins;
+import org.apache.commons.lang.StringUtils;
 
+import javax.annotation.CheckForNull;
+import javax.annotation.Nonnull;
```
jglick: Unrelated, revert.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -254,55 +254,57 @@ public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?
      * @param run the build in which the image building occurred
      */
     public static void addFromFacet(@CheckForNull String ancestorImageId, @Nonnull String descendantImageId, @Nonnull Run<?,?> run) throws IOException {
-        long timestamp = System.currentTimeMillis();
-        if (ancestorImageId != null) {
-            Fingerprint f = forImage(run, ancestorImageId);
+        synchronized (run.getParent()) {
```
jglick: The PR could be reduced to the addition of this line and its closing brace.

Anyway, this looks wrong to me. A `Fingerprint`, by design, represents something that is (potentially) shared across multiple builds, typically of different jobs. And `Fingerprint.save` synchronizes on `this`. So I suspect you rather want to be synchronizing on each `Fingerprint` being manipulated. Try:

```diff
diff --git a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
index 047b4e4..95c62f4 100644
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
@@ -257,6 +257,7 @@ public class DockerFingerprints {
         long timestamp = System.currentTimeMillis();
         if (ancestorImageId != null) {
             Fingerprint f = forImage(run, ancestorImageId);
+            synchronized (f) {
             Collection<FingerprintFacet> facets = f.getFacets();
             DockerDescendantFingerprintFacet descendantFacet = null;
             for (FingerprintFacet facet : facets) {
@@ -278,8 +279,10 @@ public class DockerFingerprints {
             } finally {
                 bc.abort();
             }
+            }
         }
         Fingerprint f = forImage(run, descendantImageId);
+        synchronized (f) {
         Collection<FingerprintFacet> facets = f.getFacets();
         DockerAncestorFingerprintFacet ancestorFacet = null;
         for (FingerprintFacet facet : facets) {
@@ -303,6 +306,7 @@ public class DockerFingerprints {
         } finally {
             bc.abort();
         }
+        }
     }
 
 }
```

and see if that fixes your issue.

bjoernhaeuser:
I removed all unrelated changes and went with the suggestion to synchronize the fingerprint. I just tested this on our jenkins installation and it looks very promising.

bjoernhaeuser:
promising as in: no concurrentmodificationexceptions anymore

bjoernhaeuser:
@jglick could you take another look please? :)

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -257,6 +257,7 @@ public static void addFromFacet(@CheckForNull String ancestorImageId, @Nonnull S
         long timestamp = System.currentTimeMillis();
         if (ancestorImageId != null) {
             Fingerprint f = forImage(run, ancestorImageId);
+            synchronized (f) {
             Collection<FingerprintFacet> facets = f.getFacets();
```
crummy: Looks like some indentation is needed here, and after the next synchronized block too

bjoernhaeuser:
This is now runnig for two weeks in our production setup and the error disappeared.

bjoernhaeuser:
@jglick Could you lift your "Requested changes"?

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
jglick: The idea was to avoid touching every line in between, which makes diffs harder to read and complicates merges & cherry-picks.

jglick:
@abayer I think you own this

abayer:
@jglick - I don't think so - I've literally never touched this plugin, don't have permissions on it, and am traveling/on vacation for the next two weeks. =)

PR #71, opened 2018-05-17
[JENKINS-51397] Define API for passing environment to docker login command

jglick:
[JENKINS-51397](https://issues.jenkins-ci.org/browse/JENKINS-51397)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
@@ -60,6 +60,7 @@ import java.util.regex.Pattern;
 
 import static com.cloudbees.plugins.credentials.CredentialsMatchers.*;
 import hudson.AbortException;
+import hudson.EnvVars;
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpoint.java
```diff
@@ -207,7 +208,7 @@ public KeyMaterialFactory newKeyMaterialFactory(Item context, @Nonnull VirtualCh
     }
 
     /**
-     * @deprecated Call {@link #newKeyMaterialFactory(Item, FilePath, Launcher, TaskListener, String)}
+     * @deprecated Call {@link #newKeyMaterialFactory(Item, FilePath, Launcher, EnvVars, TaskListener, String)}
```
dwnusbaum: nit: Maybe also a good idea to update the recommended method in the Javadoc of the two deprecated methods above this one.

jglick: Yes, good catch.

PR #74, opened 2019-04-01
[JENKINS-48437] Add new APIs to retrieve Docker credentials for a Run

dwnusbaum:
See [JENKINS-48437](https://issues.jenkins-ci.org/browse/JENKINS-48437). This subsumes #68 from @Dohbedoh with some tweaks based on [my review comments there](https://github.com/jenkinsci/docker-commons-plugin/pull/68#pullrequestreview-220710089).

Effective diff from #68 is: https://github.com/Dohbedoh/docker-commons-plugin/compare/JENKINS-48437...dwnusbaum:JENKINS-48437.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
@@ -95,7 +95,10 @@ public class DockerServerEndpoint extends AbstractDescribableImpl<DockerServerEn
     /**
      * Makes the key materials available locally for the on-going build
      * and returns {@link KeyMaterialFactory} that gives you the parameters needed to access it.
+     *
+     * @deprecated Call {@link #newKeyMaterialFactory(Run, VirtualChannel)}
      */
+    @Deprecated
     public KeyMaterialFactory newKeyMaterialFactory(@Nonnull AbstractBuild build) throws IOException, InterruptedException {
         final FilePath workspace = build.getWorkspace();
         if (workspace == null) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerEndpoint.java
```diff
@@ -100,13 +101,16 @@ public KeyMaterialFactory newKeyMaterialFactory(@Nonnull AbstractBuild build) th
         if (workspace == null) {
             throw new IllegalStateException("Build has no workspace");
         }
-        return newKeyMaterialFactory(build.getParent(), workspace.getChannel());
+        return newKeyMaterialFactory(build, workspace.getChannel());
```
dwnusbaum: This is wrong, my goal was to call the new undeprecated API directly, but we need an explicit cast to Run because as-written this is just an infinite loop.

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/DockerRegistryEndpointTest.java
```diff
@@ -111,15 +121,27 @@ public void testGetTokenForRun() throws IOException {
...
+            DockerRegistryToken token = new DockerRegistryEndpoint("https://index.docker.io/v1/", globalCredentialsId)
+                    .getToken(j.buildAndAssertSuccess(p1));
+            Assert.assertNotNull("Alice has Credentials.USE_ITEM and should be able to use the credential", token);
             Assert.assertEquals("user", token.getEmail());
             Assert.assertEquals(Base64.getEncoder().encodeToString("user:password".getBytes(Charsets.UTF_8)), token.getToken());
         }
+
+        try (ACLContext as = ACL.as(User.getById("bob", false))) {
+            DockerRegistryToken token = new DockerRegistryEndpoint("https://index.docker.io/v1/", globalCredentialsId)
+                    .getToken(j.buildAndAssertSuccess(p2));
```
jglick: Not obvious from reading this that the authentication is being applied to `getToken` and ignored by `buildAndAssertSuccess`. Would be clearer to extract the `Run` variable outside the `as` block.

dwnusbaum: Good point, done in 65ea219.

PR #75, opened 2019-04-01
Use secretTextarea for PKI certificates

jvz:
Using the backport from https://github.com/jenkinsci/jenkins/pull/3967

This replaces the textarea form input for PKI certificates with a new custom editor for multiline secrets. See the above PR for screenshots.

@jglick @batmat @dwnusbaum @reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
   </parent>
   <groupId>org.jenkins-ci.plugins</groupId>
   <artifactId>docker-commons</artifactId>
-  <version>1.15-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
   <packaging>hpi</packaging>
 
   <name>Docker Commons Plugin</name>
--- a/pom.xml
+++ b/pom.xml
@@ -10,7 +10,7 @@
   </parent>
   <groupId>org.jenkins-ci.plugins</groupId>
   <artifactId>docker-commons</artifactId>
-  <version>1.14</version>
+  <version>1.15-SNAPSHOT</version>
   <packaging>hpi</packaging>
 
   <name>Docker Commons Plugin</name>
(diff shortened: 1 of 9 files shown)


batmat:
I do not understand how this can use `secretTextarea` without actually bumping to Jenkins 2.171+, which is where https://github.com/jenkinsci/jenkins/pull/3967 landed?

batmat:
OK, answering to self, https://github.com/jenkinsci/lib-multiline-secrets-ui is providing this. But then I'm still a bit lost why this was both added to the Core and https://github.com/jenkinsci/lib-multiline-secrets-ui

jvz:
The standalone library is temporary for use during the time that the core feature isn't part of an LTS release yet. Then we can bump the various plugins to require that LTS release and drop the library.

pom.xml
```diff
@@ -67,6 +67,11 @@
       <artifactId>authentication-tokens</artifactId>
       <version>1.3</version>
     </dependency>
+    <dependency>
+      <groupId>io.jenkins.temp.jelly</groupId>
+      <artifactId>multiline-secrets-ui</artifactId>
```
jglick: This is in 2.171 so why not simply adjust `jenkins.version` and skip this library?

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
```diff
@@ -70,6 +72,12 @@ public String getClientKey() {
         return clientKey == null ? null : clientKey.getPlainText();
     }
 
+    @CheckForNull
+    @Restricted(NoExternalUse.class)
+    public Secret getClientKeySecret() {
```
jglick: IMO better to just change the type of the original getter and constructor parameter. AFAIK they are not referred to from other plugins. (From a _test_ in `docker-workflow` but we do not care about that.)

pom.xml
jvz: Can do. This PR predates the PR that was merged into 2.171.

jglick: Right, that was just released I guess.

jglick:
Compilation error it seems.

jvz:
> Looks OK. Did you check binary compatibility with `docker-workflow`?

Not yet. I'll verify that.

jvz:
Messed up the history, re-creating this PR.

PR #76, opened 2019-04-22
[JENKINS-57420] - Use f:secretTextarea for client private key

jvz:
This updates the form UI for client keys to use the newly introduced
f:secretTextarea Jelly tag from Jenkins 2.171. Now the client key can
be more securely entered and updated.

Recreated based on #75.

@jglick @jeffret-b @Wadeck @dwnusbaum

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.42</version>
+    <version>3.43</version>
     <relativePath />
   </parent>
   <groupId>org.jenkins-ci.plugins</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -33,7 +33,7 @@
   <properties>
     <revision>1.15</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.171</jenkins.version>
+    <jenkins.version>2.60.3</jenkins.version>
     <java.level>8</java.level>
   </properties>
 
@@ -66,6 +66,12 @@
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>authentication-tokens</artifactId>
       <version>1.3</version>
+    </dependency>
+    <!-- TODO: remove after upgrading to Jenkins 2.171+ and migrating to f:secretTextarea -->
+    <dependency>
+      <groupId>io.jenkins.temp.jelly</groupId>
+      <artifactId>multiline-secrets-ui</artifactId>
+      <version>1.0</version>
     </dependency>
 
     <!-- for Pipeline-based unit tests -->
--- a/pom.xml
+++ b/pom.xml
@@ -5,12 +5,12 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.12</version>
+    <version>3.42</version>
     <relativePath />
   </parent>
   <groupId>org.jenkins-ci.plugins</groupId>
   <artifactId>docker-commons</artifactId>
-  <version>1.15-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
   <packaging>hpi</packaging>
 
   <name>Docker Commons Plugin</name>
@@ -31,9 +31,9 @@
   </scm>
 
   <properties>
-    <revision>1.14</revision>
+    <revision>1.15</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.171</jenkins.version>
     <java.level>8</java.level>
   </properties>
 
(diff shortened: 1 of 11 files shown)


jvz:
@jglick I tested this with docker-workflow-plugin, and it still compiles fine. I can make a test PR with the incremental release when it's uploaded.

jvz:
@oleg-nenashev

oleg-nenashev:
I will review it and release tomorrow if it is feasible

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
```diff
@@ -66,8 +74,8 @@ public DockerServerCredentials(CredentialsScope scope, String id, String descrip
      * @return null if there's no authentication
      */
     @CheckForNull
-    public String getClientKey() {
-        return clientKey == null ? null : clientKey.getPlainText();
+    public Secret getClientKey() {
```
oleg-nenashev: It breaks binary compatibility.
Usage example: https://github.com/jenkinsci/docker-swarm-plugin/blob/aab115b35c59758ad6fc38ca01fee1f11fbfa2a0/src/main/java/org/jenkinsci/plugins/docker/swarm/DockerServerCredentialsSSLConfig.java#L29

jvz: What is this CertificateUtils? Not part of docker-commons.

jvz: Also, the alternative is to leave the getter as is and break data bound constructor.

jglick: Do not break databinding please. You could leave the existing getter, deprecated, and introduce a new property with a different name.

jvz: By renaming the property, we get a different JCasC syntax potentially. Or am I confusing something?

jvz: I'm renaming the property to `clientKeySecret` while leaving the constructor and getter for ABI compatibility. Those are now deprecated.

jvz:
@oleg-nenashev I've addressed the ABI compatibility issue.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
```diff
@@ -50,24 +50,42 @@
...
-        this.clientKey = Util.fixEmptyAndTrim(clientKey) == null ? null : Secret.fromString(clientKey);
+        this.clientKey = clientKeySecret;
         this.clientCertificate = Util.fixEmptyAndTrim(clientCertificate);
         this.serverCaCertificate = Util.fixEmptyAndTrim(serverCaCertificate);
     }
 
+    /**
+     * @deprecated use {@link #getClientKeySecret()}
+     */
+    @CheckForNull
```
oleg-nenashev:
```suggestion
    @CheckForNull
    @Deprecated
```

pom.xml
```diff
@@ -5,12 +5,12 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.12</version>
+    <version>3.42</version>
```
jglick:
```suggestion
    <version>3.43</version>
```

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentials.java
```diff
@@ -50,24 +50,42 @@
     @CheckForNull 
     private final String serverCaCertificate;
 
-    @DataBoundConstructor
+    /**
+     * @deprecated use {@link #DockerServerCredentials(CredentialsScope, String, String, Secret, String, String)}
+     */
     public DockerServerCredentials(CredentialsScope scope, String id, String description,
```
jglick:
```suggestion
    @Deprecated
    public DockerServerCredentials(CredentialsScope scope, String id, String description,
```

jvz:
@oleg-nenashev can this be merged now?

pom.xml
```diff
@@ -31,9 +31,9 @@
   </scm>
 
   <properties>
-    <revision>1.14</revision>
+    <revision>1.15</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.171</jenkins.version>
```
oleg-nenashev: Would be better to support recent LTS versions

jvz:
@oleg-nenashev I've reverted the version upgrade and am now using the temporary library instead.

oleg-nenashev:
https://issues.jenkins-ci.org/browse/JENKINS-57420 ftr

PR #77, opened 2019-04-30
[JENKINS-57439 ] - Add declarative support

lion7:
This PR implements the credentials handler for a declarative pipeline, similiar to https://github.com/jenkinsci/aws-credentials-plugin/pull/50 and inspired by [FileCredentialsHandler.java](https://github.com/jenkinsci/pipeline-model-definition-plugin/blob/master/pipeline-model-definition/src/main/java/org/jenkinsci/plugins/pipeline/modeldefinition/credentials/impl/FileCredentialsHandler.java)
The dependency on `pipeline-model-extensions` is made optional as suggested by @jglick in https://github.com/jenkinsci/aws-credentials-plugin/pull/50,

Additionally I also added a symbol annotation to the `DockerTool` descriptor so it can be used in the tools section of a declarative pipeline (inspired by [GradleInstallation.java](https://github.com/jenkinsci/gradle-plugin/blob/master/src/main/java/hudson/plugins/gradle/GradleInstallation.java#L95))

This allows to use the Docker tool and Docker credentials like this:
```
environment {
  DOCKER_CERT_PATH = credentials('id-for-a-docker-cred')
}
tools {
  docker '18.09'
}
stages {
  stage('foo') {
    steps {
      sh "docker version" // DOCKER_CERT_PATH is automatically picked up by the Docker client
    }
  }
}
```

Note that I had to bump a lot of dependency versions due to the enforcer plugin. I hope this is ok.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,6 +77,7 @@
       <version>1.3.8</version>
       <!-- optional because the dependency subtree is quite large -->
       <optional>true</optional>
+    </dependency>
     <!-- TODO: remove after upgrading to Jenkins 2.171+ and migrating to f:secretTextarea -->
     <dependency>
       <groupId>io.jenkins.temp.jelly</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -155,6 +155,37 @@
       <classifier>tests</classifier>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jenkinsci.plugins</groupId>
+      <artifactId>pipeline-model-definition</artifactId>
+      <version>1.3.8</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- these dependencies are added to satisfy the enforcer rules -->
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-lang3</artifactId>
+      <version>3.7</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>mailer</artifactId>
+      <version>1.20</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins</groupId>
+      <artifactId>ssh-credentials</artifactId>
+      <version>1.13</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jenkins-ci.plugins.workflow</groupId>
+      <artifactId>workflow-scm-step</artifactId>
+      <version>2.6</version>
+      <scope>test</scope>
+    </dependency>
 
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -5,9 +5,10 @@
   <parent>
     <groupId>org.jenkins-ci.plugins</groupId>
     <artifactId>plugin</artifactId>
-    <version>3.12</version>
+    <version>3.42</version>
     <relativePath />
   </parent>
+
   <groupId>org.jenkins-ci.plugins</groupId>
   <artifactId>docker-commons</artifactId>
   <version>${revision}${changelist}</version>
@@ -35,7 +36,7 @@
     <changelist>-SNAPSHOT</changelist>
     <revision>1.15</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.164.2</jenkins.version>
     <java.level>8</java.level>
   </properties>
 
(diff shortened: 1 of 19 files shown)


lion7:
What's the usual proces for getting a PR like this reviewed? I see that most PRs are usually reviewed by @jglick, @jvz or @dwnusbaum?

pom.xml
```diff
@@ -55,92 +58,99 @@
...
       <groupId>org.jenkins-ci.plugins</groupId>
       <artifactId>authentication-tokens</artifactId>
       <version>1.3</version>
     </dependency>
+    <dependency>
+      <groupId>org.jenkinsci.plugins</groupId>
+      <artifactId>pipeline-model-extensions</artifactId>
+      <version>1.3.8</version>
+      <!-- optional because the dependency subtree is quite large -->
+      <optional>true</optional>
```
dwnusbaum: Fair enough (for reasons similar to https://github.com/jenkinsci/aws-credentials-plugin/pull/50#discussion_r271934841), but you need to change `@Extensions` that use the optional dependency to specify `optional = true` or the plugin will fail at runtime when the optional dependency is not present. In unit/integration tests here, optional dependencies are installed, so it is easy to get this wrong, which is one of the reasons optional dependencies are painful.

pom.xml
```diff
@@ -31,9 +32,11 @@
   </scm>
 
   <properties>
-    <revision>1.14</revision>
+    <revision>1.15</revision>
+    <changelist>-SNAPSHOT</changelist>
+    <revision>1.15</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.60.3</jenkins.version>
+    <jenkins.version>2.164.2</jenkins.version>
```
dwnusbaum: Any particular reason to bump to the latest LTS? If not, I'd prefer to update to 2.138.4 at most since I think that is the oldest LTS line still getting security updates.

pom.xml
```diff
@@ -31,9 +32,11 @@
   </scm>
 
   <properties>
-    <revision>1.14</revision>
+    <revision>1.15</revision>
+    <changelist>-SNAPSHOT</changelist>
```
dwnusbaum: Looks like these properties are dupes.

pom.xml
```diff
@@ -5,12 +5,13 @@
...
     <artifactId>plugin</artifactId>
-    <version>3.12</version>
+    <version>3.42</version>
     <relativePath />
   </parent>
+
   <groupId>org.jenkins-ci.plugins</groupId>
   <artifactId>docker-commons</artifactId>
-  <version>1.15-SNAPSHOT</version>
+  <version>${revision}${changelist}</version>
```
dwnusbaum: Ah whoops, I should have reincrementalified after the last release because the parent pom is older than the version where that start happening automatically (3.18), thanks for fixing it!

pom.xml
```diff
@@ -55,92 +58,99 @@
...
       <artifactId>workflow-cps</artifactId>
-      <version>2.39</version>
+      <version>2.67</version>
       <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>org.jenkins-ci.plugins.workflow</groupId>
       <artifactId>workflow-cps</artifactId>
-      <version>2.39</version>
+      <version>2.67</version>
```
dwnusbaum: nit: It would be good to introduce properties for the versions of workflow-step-api, workflow-cps, and workflow-support, since they are each used twice and we don't want them to get out of sync.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsHandler.java
```diff
@@ -0,0 +1,61 @@
...
+import org.jenkinsci.plugins.credentialsbinding.MultiBinding;
+import org.jenkinsci.plugins.pipeline.modeldefinition.model.CredentialsBindingHandler;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Extension
```
dwnusbaum: Needs `optional = true`.

pom.xml
lion7: That is good to know, I marked the `@Extension` annotation on `DockerServerCredentialsHandler` as optional

pom.xml
lion7: Not really, the latest LTS just felt like a safe bet. I changed it to 2.138.4.

pom.xml
lion7: You are correct, this was something that slipped through when merging

pom.xml
lion7: Your welcome, but you should really thank the enforcer plugin for nagging about it

pom.xml
lion7: Good suggestion, I created a few properties for those deps

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/DockerServerCredentialsHandler.java
lion7: Done

lion7:
Thanks for your quick review @dwnusbaum 😄

pom.xml
dwnusbaum: Thanks! Minor nits but `workflow-step-cps` and `workflow-step-support` should just be `workflow-cps` and `workflow-support`, so it would be great to go ahead and fix that.

lion7:
Whoops, committing without running a local build is a bad idea 🙈 

Anyway, the lowest compatible Jenkins version seems to be 2.150.1.

lion7:
By the way, the build fails with a `No space left on device`...

abayer:
I've kicked off a new build which should get a new agent.

lion7:
> I've kicked off a new build which should get a new agent.

The build still fails with the same error 😢

abayer:
There was one agent that just didn’t want to go away despite being out of disk space - I’ve nuked it by hand and rerun.

lion7:
My next (probably obvious) question: how are things normally merged? Does a contributor merge this PR? And how/when are changes like this usually released and made available in the Jenkins plugin portal? 
Currently I installed a locally build version of this plugin in our corporate Jenkins but the IT support team isn't really happy with customized plugins (both from a maintenance and security perspective) seeing that they manage a dozen instances of Jenkins.

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerTool.java
```diff
@@ -114,7 +115,8 @@ public DockerTool forNode(Node node, TaskListener log) throws IOException, Inter
         }
     }
 
-    @Extension public static class DescriptorImpl extends ToolDescriptor<DockerTool> {
+    @Extension @Symbol("docker")
```
jglick: This should be a separate PR. See #52.

lion7: Agreed, especially after reading the discussion in #52 . It looked like a simple change so I just included it in this PR. But I will move it to a separate PR.

lion7:
I'll try to add a test that tests the new class explicitly instead of relying on `DockerServerCredentialsBindingTest`. 

As for whether this change is really necessary: it depends I geuss. In our workflow we try to stick as much as possible to declarative pipelines to keep things simple and understandable (writing this comment reminds me of https://xkcd.com/1172/).

Its true that we could use `withDockerServer` (actually `docker.withServer()` since the step itself shouldn't be used directly) but that means I have to repeat this in every stage and also I'd probably have to use a script block since `docker.with***()` refers to a global variable and not a step. IMO that kind of defeats the whole purpose of using a declarative pipeline, in that case I might as well use a scripted pipeline instead.

What I'm trying to archieve is running a multi-stage build on a Docker EE UCP cluster. Most of our build tool logic (we're using Gradle) is inside the `Dockerfile`, so basicly Jenkins only has to run a `docker build`, passing a few credentials via build arguments and then push the resulting image to our corporate Docker registry. I'll try to think of a few more alternative next week to figure out if I can archieve a working pipeline without bringing in too much complexity in our pipeline code.

jglick:
The `docker` global variable is incompatible with Declarative and IMHO should never have been created to begin with.  `withDockerServer` is fine and should work in Declarative. But yes it would need to be used separately in each `stage` requiring access to the Docker daemon. (Though I thought I remembered some provision in Declarative by which a block-scoped step could be used automatically in an `options` block?)

To be clear, I am not trying to block this PR, just clarify the use case vs. alternatives.

lion7:
Hmm, the only steps that seem to be usable in an `options` block are the `retry` and `timestamps` steps. I'm not entirely sure how that works, but if the `withDockerServer` step could be used in an `options` block then this PR isn't that necessary anymore since that would enable me to define the Docker daemon to use for the entire pipeline or (set of) stage(s). I geuss the same goes for the `withDockerRegistry` step.

No offense taken in your earlier comment btw, I totally agree that it's better to discuss alternatives before blindly merging changes. 

On a sidenote, if the `docker` global variable is a bad idea then maybe it should be promoted less in the documentation? Currently its use is encouraged and the use of the individual steps are discouraged.

lion7:
I added a test for the new functionality. I did have to add some dependencies to allow the usage of a declarative pipeline in the test code. 

As a practical alternative I attached our UCP manager node as an agent to Jenkins. This way both the `docker` command and Docker certificates are available locally on the agent, so in my pipeline I only have to specify the label for this specific agent. The downside of this is that a part of the pipeline configuration is now 'hidden' in the node configuration. We have yet to discuss internally if that is acceptable or that we'd rather be explicit about this configuration in the pipeline code.

lion7:
Are there any plans to merge this soon? Just hoping this PR won't die a silent death 😛

oleg-nenashev:
resolved the conflict. Will review ASAP and merge if everything is fine from my PoV

lion7:
Thanks @oleg-nenashev  😄

oleg-nenashev:
@lion7 is there a Jenkins JIRA ticket for it so that I reference it in the changelog?

oleg-nenashev:
@lion7 my plan is to cut the release today unless there is a negative feedback.
Would it make sense to add a Pipeline support section to the README.md and to add the sample there. Some documentation in the plugin would not hurt for sure

lion7:
@oleg-nenashev I added the example in this PR to the README. 
There is no JIRA issue yet, and I can't create a JIRA account because the email containing the password won't show up in my inbox 😞

oleg-nenashev:
Created https://issues.jenkins-ci.org/browse/JENKINS-57439 . Not sure what causes the JIRA access issue tho, maybe it makes sense to contact the Jenkins INFRA team

PR #79, opened 2019-06-27
Make the credentials label more generic

Vlatombe:
Despite initially introduced to provide authentication to a Docker host, this
credentials format is actually independant from Docker itself, and can be used
to authenticate to a TLS server using a X.509 client certificate.

So this adjusts the label accordingly so that users can discover it in an
easier way for other use cases than Docker (such as Kubernetes). See downstream usage in https://github.com/jenkinsci/kubernetes-plugin/pull/520


oleg-nenashev:
@Vlatombe I will release that, but it might be useful to move it to the Credentials plugin later. It is a common functionality after all

PR #85, opened 2020-01-08
Add ability to disable fingerprint recording

jonsten:
It seems like the fingerprinting functionality has been broken for a while (e.g. [see this comment by Jesse Glick](https://issues.jenkins-ci.org/browse/JENKINS-46498?focusedCommentId=338220#comment-338220)) and causes a lot of problems, especially in large Jenkins instances. This change introduces a property which disables the functionality all together.

Some background, in our case, where we run 10k+ builds a day using the same docker image, the fingerprint file for that image grows very large 50+ MB and around 100k container entries. The large size causes Jenkins to spend more and more time saving the file (serialization, file IO, etc). In the end, when it gets large enough, the builds are aborted without anything being written in the logs (separate issue, but it seems like if more than five minutes is spent in the step start / fingerprint recording phase, then the build is terminated without anything written in the log). At that size it takes around 3-4 seconds to save the file, and if 100 builds are starting containers at the same time, then some of them will spend more than five minutes waiting.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
@@ -23,6 +23,7 @@
  */
 package org.jenkinsci.plugins.docker.commons.fingerprint;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.BulkChange;
 import hudson.model.Fingerprint;
 import hudson.model.Run;
@@ -47,6 +48,9 @@ import org.apache.commons.lang.StringUtils;
 public class DockerFingerprints {
     
     private static final Logger LOGGER = Logger.getLogger(DockerFingerprints.class.getName());
+
+    @SuppressFBWarnings(value="MS_SHOULD_BE_FINAL", justification="mutable for scripts")
+    public static boolean DISABLE = Boolean.getBoolean(DockerFingerprints.class.getName() + ".DISABLE");
     
     private DockerFingerprints() {} // no instantiation
  
@@ -221,6 +225,9 @@ public class DockerFingerprints {
      * Adds a new {@link ContainerRecord} for the specified image, creating necessary intermediate objects as it goes.
      */
     public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?> run) throws IOException {
+        if (DISABLE) {
+            return;
+        }
         String imageId = record.getImageId();
         Fingerprint f = forImage(run, imageId);
         synchronized (f) {
@@ -256,6 +263,9 @@ public class DockerFingerprints {
      * @param run the build in which the image building occurred
      */
     public static void addFromFacet(@CheckForNull String ancestorImageId, @Nonnull String descendantImageId, @Nonnull Run<?,?> run) throws IOException {
+        if (DISABLE) {
+            return;
+        }
         long timestamp = System.currentTimeMillis();
         if (ancestorImageId != null) {
             Fingerprint f = forImage(run, ancestorImageId);


src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -47,6 +48,9 @@
 public class DockerFingerprints {
     
     private static final Logger LOGGER = Logger.getLogger(DockerFingerprints.class.getName());
+
+    @SuppressFBWarnings(value="MS_SHOULD_BE_FINAL", justification="mutable for scripts")
+    public static boolean DISABLE = Boolean.getBoolean(DockerFingerprints.class.getName() + ".DISABLE");
```
oleg-nenashev: Add `@Restricted(NoExternalUse.class)` and add a static getter method to offer API for downstream plugins which may also need this flag?

jonsten:
Thank you for the quick feedback!

I opted for the system property as it is the "quick" solution and it doesn't bloat system configuration with yet another checkbox. With that said I agree with you that this might be a rather useful setting. In fact, in a perfect world it should be possible to set from case to case, e.g. same way as it is done for artifacts. Unfortunately I'm not willing to spend the effort to implement all of that. I will see if there is a natural place to add a system config, if not I will go with the system property route.

> I am requesting changes, because the new option needs to be documented in README.

Definitely I'll fix!

>Javadoc for the new property

I'll fix!

>New Static Getter method. It will be useful for Dwnstream plugins, because they can optimize their logic when fingerprint recording is disabled

Shouldn't they be able to use the field directly or that is a no-go way to solve it?

oleg-nenashev:
> Shouldn't they be able to use the field directly or that is a no-go way to solve it?

It is also possible, but it will complicate further changes should we decide to change/improve behavior later (e.g. adding a global configuration). Static methods are more convenient from the extensibility PoV

jonsten:
> > Shouldn't they be able to use the field directly or that is a no-go way to solve it?
> 
> It is also possible, but it will complicate further changes should we decide to change/improve behavior later (e.g. adding a global configuration). Static methods are more convenient from the extensibility PoV

Got ya, I'll add a static method!

As for global config, I see no natural home for such an option, so, let's start with the property for now. Adding information in the README and in code should be enough for people to find it if they start to run into performance problems.

src/main/java/org/jenkinsci/plugins/docker/commons/fingerprint/DockerFingerprints.java
```diff
@@ -221,6 +232,9 @@ private DockerFingerprints() {} // no instantiation
      * Adds a new {@link ContainerRecord} for the specified image, creating necessary intermediate objects as it goes.
      */
     public static void addRunFacet(@Nonnull ContainerRecord record, @Nonnull Run<?,?> run) throws IOException {
+        if (DISABLE) {
+            return;
+        }
```
oleg-nenashev: Would it also make sense to add logging on the FINE or INFO levels?

jonsten: I'm not the expert here, after all, I'm the one making a change to disable all of this, so I have a hard time finding a point where it would be useful. But if you can see a reason for adding it, I'll happily add it :) Besides, at level FINE/INFO it would require the user to actively set the log level in order to get it to the system logs etc (have it on by default would be way to chatty IMHO).

jonsten:
> Should no longer be necessary as of [jenkinsci/docker-workflow-plugin#180](https://github.com/jenkinsci/docker-workflow-plugin/pull/180) in 1.19 plus [jenkinsci/pipeline-model-definition-plugin#350](https://github.com/jenkinsci/pipeline-model-definition-plugin/pull/350) in 1.4.0.

Ummm okay, I've might have missed something, so please explain how I prevent fingerprint recording without this change when i run a scripted pipeline, e.g.:

    def image = docker.image('some_image')
    image.inside {
        ...
    }

With the latest release of docker common and docker workflow plugin this code will record fingerprints.

jglick:
Yes https://github.com/jenkinsci/docker-workflow-plugin/blob/74a2370901f41e8b5b541d768b440e2ab1cd1b18/src/main/java/org/jenkinsci/plugins/docker/workflow/WithContainerStep.java#L209 was forgotten, as were a couple lines in `DockerDSL/help.jelly`.

jonsten:
Okay, then it sounds like I could scrap this change and instead remove the call to addRunFacet in WithContainerStep. I'll see what I can do tomorrow.

oleg-nenashev:
@jglick @jonsten I think this is still relevant, because there are other plugins relying on Fingerprints. E.g. DockerHub Notifications IIRC. Pipeline is not the only API where it is used. One of the follow-ups for the Docker Traceability project in 2015 was to improve fingerprints performance, but it has never happened as you may guess :( https://jenkins.io/projects/gsoc/2020/project-ideas/external-fingerprint-storage-for-jenkins/ in GSoC may somewhat address it

jglick:
I do not think other plugins _produce_ Docker fingerprints. Anyway the request here was about `docker-workflow`.

jonsten:
Based on what @jglick bought up I've created https://github.com/jenkinsci/docker-workflow-plugin/pull/198 over at docker-workflow. I've also added the log entries you suggested @oleg-nenashev. I leave it to you guys to decide which one to merge (or both(, or none)).

PR #87, opened 2020-06-30
Configuration as Code support

rsandell:
Added tests to verify it works.
Added Symbols to make the configuration more neat looking and readable.
Updated pom to use the current JCasC version.


pom.xml
```diff
@@ -34,11 +34,10 @@
...
     <revision>1.17</revision>
     <changelist>-SNAPSHOT</changelist>
-    <jenkins.version>2.150.1</jenkins.version>
+    <jenkins.version>2.222.4</jenkins.version>
     <java.level>8</java.level>
-    <workflow-step-api.version>2.19</workflow-step-api.version>
-    <workflow-cps.version>2.67</workflow-cps.version>
-    <workflow-support.version>3.3</workflow-support.version>
+    <pipeline-model-definition-version>1.3.8</pipeline-model-definition-version>
+    <jcasc.version>1.41</jcasc.version>
```
alecharp, line 40: as you are using the plugin BOM (which is great!), the JCasC version is normally defined in it. Is it not?

src/test/resources/org/jenkinsci/plugins/docker/commons/casc_bare.yaml
```diff
@@ -0,0 +1,22 @@
...
+  dockerTool:
+    installations:
+      - name: "docker-latest"
+        properties:
+          - installSource:
+              installers:
+                - docker:
+                    version: "latest"
+      - home: "/etc/docket/docker"
+        name: "docker-native"
```
alecharp, line 22: Non mandatory suggestion

```suggestion
        name: "docker-native"

```

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerTool.java
```diff
@@ -114,7 +115,8 @@ public DockerTool forNode(Node node, TaskListener log) throws IOException, Inter
         }
     }
 
-    @Extension public static class DescriptorImpl extends ToolDescriptor<DockerTool> {
+    @Extension @Symbol("dockerTool")
```
varyvol, line 118: It was already the symbol used, so I'm not sure adding it is confusing?

rsandell: Just in case the algorithm for the default name changes in the future

pom.xml
rsandell: Not sure, a bunch of stuff isn't

alecharp: it seems that you can. but it's not a blocker for me

src/main/java/org/jenkinsci/plugins/docker/commons/tools/DockerTool.java
jglick: https://issues.jenkins.io/browse/JENKINS-37390

PR #90, opened 2021-04-23
Install docker from stable channel

jinogueiraa:
Edge channel has 19.03.9 as latest version available
https://download.docker.com/linux/static/edge/x86_64/
Imho it has to be changed  to stable channel
https://download.docker.com/linux/static/stable/x86_64/

- [x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!
- [x] Ensure that the pull request title represents the desired changelog entry
- [x] Please describe what you did
- [x] Link to relevant issues in GitHub or Jira (none found)
- [x] Link to relevant pull requests, esp. upstream and downstream changes (none found)
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

PR #92, opened 2021-07-28
Update DockerToolInstaller.java to use stable instead of edge releases

jonl-percsolutions-com:
Updated the url for docker installs from download.docker.com to use the stable releases instead of edge.


nfalco79:
I've got the same issue. stable path should be use in place of edge.

marciejones:
I'm having the same issue (Jenkins cannot install Docker due to it using the wrong download path).  Example:  Failed to install https://download.docker.com/win/static/edge/x86_64/docker-20.10.6.tgz

Should use _stable_:  https://download.docker.com/win/static/**stable**/x86_64/

I'd love to see this fixed/merged.

jinogueiraa:
There is another pull request that addresses this issue:  #90 (Yes, i did it)

Please take a look and comment. (And consider merge if appropriate and you have rigths to do it...)

jinogueiraa:
Duplicate of #90

PR #93, opened 2022-01-12
[JENKINS-67572] Allow docker digest in image names

viceice:
Allow docker digests in image names, as since c069b79c31c5aa80a01b0c462f0dc6b41751f059 it's no longer allowed in `docker-workflow-plugin` to pin docker images.

- fixes https://issues.jenkins.io/browse/JENKINS-67572

- [x] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue


PandarinDev:
Great job, looking forward to this getting merged! Currently our workflows are failing because of this regression and we cannot really downgrade our plugins so hoping that this will be released ASAP.

j3t:
As a workaround, you can disable the validation see https://github.com/jenkinsci/docker-commons-plugin/blob/master/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java#L48

viceice:
@j3t I know, but that fully disables all checks again and we are vulnerable again. So i hope this gets merge and released soon.

@rsandell @oleg-nenashev Any changes to get this done sooner than later?

jsnod:
> As a workaround, you can disable the validation see https://github.com/jenkinsci/docker-commons-plugin/blob/master/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java#L48

@j3t We are starting Jenkins with `-Dorg.jenkinsci.plugins.docker.commons.credentials.ImageNameValidator.SKIP=true` but we are still failing on validation, is there some other way to enable this `SKIP` workaround?

viceice:
I think you are probably also hit by

- https://github.com/jenkinsci/docker-workflow-plugin/pull/244

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
```diff
@@ -129,6 +147,43 @@ public static void checkUserAndRepo(@NonNull String userAndRepo) throws FormVali
...
+    /**
+     * A digest starts with 'sha256:' and must be valid ASCII and may contain
+     * lowercase and digits.
+     * A digest contains 71 additional characters.
+     *
+     * @see <a href=
+     *      "https://docs.docker.com/engine/reference/commandline/images/#list-the-full-length-image-ids">docker
+     *      digests</a>
+     */
+    public static final Pattern VALID_DIGEST = Pattern.compile("^sha256:([a-z0-9]){64}$");
```
TBBle, line 159: Should we also allow other digest forms supported by the [OCI Image spec](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#digests)?

```
digest                ::= algorithm ":" encoded
algorithm             ::= algorithm-component (algorithm-separator algorithm-component)*
algorithm-component   ::= [a-z0-9]+
algorithm-separator   ::= [+._-]
encoded               ::= [a-zA-Z0-9=_-]+
```

`sha256` and `sha512` are registered and as-such have [further validation rules](https://github.com/opencontainers/image-spec/blob/v1.0.1/descriptor.md#registered-algorithms)

It might be better to use either that link, or https://docs.docker.com/registry/spec/api/#content-digests (the predecessor to [the OCI Distribution spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#definitions) which uses the OCI Image spec digest rules I linked above), since the current links only shows _examples_ of valid digests but does not define the validation rule.

The Docker Registry V2's rule accepts more than the OCI sha256 rule, and is neither a strict subset nor superset of the OCI general algorithm rule, so I think we can't expect to get a validation rule that doesn't still carry the risk of being rejected by the registry being talked to.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
```diff
@@ -99,21 +111,27 @@ public static boolean skipped() {
             return FormValidation.ok();
         }
         final String[] args = splitUserAndRepo(userAndRepo);
-        if (StringUtils.isBlank(args[0]) && StringUtils.isBlank(args[1]) && StringUtils.isBlank(args[2])) {
+        if (StringUtils.isBlank(args[0]) && StringUtils.isBlank(args[1]) && StringUtils.isBlank(args[2])
+                && StringUtils.isBlank(args[3])) {
```
rsandell, line 115: I think it should probably be
```suggestion
        if (StringUtils.isBlank(args[0]) && StringUtils.isBlank(args[1]) && (StringUtils.isBlank(args[2])
                || StringUtils.isBlank(args[3]))) {
```
?

src/test/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidatorTest.java
```diff
@@ -15,30 +15,36 @@
...
+                {"workflow-demo",                                                                                                    FormValidation.Kind.OK},
+                {"workflow-demo:latest@sha256:56930391cf0e1be83108422bbef43001650cfb75f64b3429928f0c5986fdb750",                     FormValidation.Kind.OK},
+                {"workflow-demo:latest@sha256:56930391cf0e1be83108422bbef43001650cfb75f64b",                                         FormValidation.Kind.ERROR},
+                {"workflow-demo@sha256:56930391cf0e1be83108422bbef43001650cfb75f64b3429928f0c5986fdb750",                            FormValidation.Kind.OK},
+                {"jenkinsci/workflow-demo@sha256:56930391cf0e1be83108422bbef43001650cfb75f64b3429928f0c5986fdb750",                  FormValidation.Kind.OK},
+                {"docker:80/jenkinsci/workflow-demo@sha256:56930391cf0e1be83108422bbef43001650cfb75f64b3429928f0c5986fdb750",        FormValidation.Kind.OK},
+                {"docker:80/jenkinsci/workflow-demo:latest@sha256:56930391cf0e1be83108422bbef43001650cfb75f64b3429928f0c5986fdb750", FormValidation.Kind.OK},
```
rsandell, line 29:
```suggestion
                {"docker:80/jenkinsci/workflow-demo:latest@sha256:56930391cf0e1be83108422bbef43001650cfb75f64b3429928f0c5986fdb750", FormValidation.Kind.OK},
                {"docker:80/jenkinsci/workflow-demo:latest@sha1:0123456789abcdef", FormValidation.Kind.ERROR},                                //?
                {"workflow-demo:latest@",                                         FormValidation.Kind.ERROR},
                {"workflow-demo@",                            FormValidation.Kind.ERROR},
                {"jenkinsci/workflow-demo@",                  FormValidation.Kind.ERROR},
                {"docker:80/jenkinsci/workflow-demo@",        FormValidation.Kind.ERROR},
                {"docker:80/jenkinsci/workflow-demo:latest@", FormValidation.Kind.ERROR},
```

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
rsandell: But the tests passes so maybe not :)

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
viceice: OK, which one to implement? I like to get this done because of the currently breaking issue? Probably the OCI spec ist best.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
viceice: not needed 🙃

viceice:
OK, fixed tests and now validating againd oci spec. So please review again. 🤗

rsandell (approved):

viceice:
♥️

viceice:
Thanks, updated jenkins and removed workaround and everything works 🎉

ccayg-sainsburys:
Is this possibly fixing some cases but not all or has something here / nearby changed that means my configuration no longer works?

I'm seeing: `ERROR: Name must follow the pattern &#039;^[a-zA-Z0-9]+((\.|_|__|-+)[a-zA-Z0-9]+)*$&#039;`
following `DOCKER_IMAGE = docker.build('${ECR_REGISTRY}/${ECR_REPO}:${COMMIT_HASH}')`

Where the registry variable is pulled from credentials while the repo and hash are defined within the pipeline itself.
This should equate to what I believe is a normal ECR reference per https://docs.aws.amazon.com/AmazonECR/latest/userguide/docker-push-ecr-image.html of the form `123456789012.dkr.ecr.eu-west-1.amazonaws.com/service-foo:0a1b23c`

with v1.19 of this plugin and v1.28 of docker-workflow

TBBle:
@ccayg-sainsburys Yeah, this is almost certainly [JENKINS-67633](https://issues.jenkins.io/browse/JENKINS-67633). I've left a comment there so it's more-visible to users with that issue.

ccayg-sainsburys:
> @ccayg-sainsburys Yeah, this is almost certainly [JENKINS-67633](https://issues.jenkins.io/browse/JENKINS-67633). I've left a comment there so it's more-visible to users with that issue.

Thanks - and the suggested fix there does work but because one of the values comes from credentials it triggers:
```
10:33:27  Warning: A secret was passed to "withEnv" using Groovy String interpolation, which is insecure.
10:33:27  		 Affected argument(s) used the following variable(s): [ECR_REGISTRY]
10:33:27  		 See https://jenkins.io/redirect/groovy-string-interpolation for details.

```
I might just replace with shell commands at this point to be honest.

TBBle:
For data from credentials, don't use `withEnv`, use the specialised credential-handling step [`withCredentials` to expose the value as an env-var](https://plugins.jenkins.io/credentials-binding/). That should fix that warning.

PR #94, opened 2022-02-13
Allow to use registry as host and port only to allow to use podman as executable

mswiderski:
- [x] Link to relevant issues in GitHub or Jira
- [ ] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests - that demonstrates feature works or fixes the issue

This is a fix for docker workflow plugin issue that manifests when using `podman`. The fix here allows to either directly (based on tool name) or via environment variable to use just how and port of the registry url instead of complete one.
https://issues.jenkins.io/browse/JENKINS-67418

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
@@ -52,6 +52,8 @@ public class RegistryKeyMaterialFactory extends KeyMaterialFactory {
 
     private static final String DOCKER_CONFIG_FILENAME = "config.json";
     private static final String[] BLACKLISTED_PROPERTIES = { "auths", "credsStore" };
+    
+    protected static final String DOCKER_REGISTRY_HOST_ONLY = "DOCKER_REGISTRY_HOST_ONLY";
 
     private final @Nonnull String username;
     private final @Nonnull String password;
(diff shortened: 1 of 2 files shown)


mmusenbr:
Is there anything to support to bring this PR forward? We currently face this issue as the infrastructure provided uses now podman for agents.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
```diff
@@ -111,6 +113,14 @@ public KeyMaterial materialize() throws IOException, InterruptedException {
         }
         return new RegistryKeyMaterial(dockerConfig, new EnvVars("DOCKER_CONFIG", dockerConfig.getRemote()));
     }
+    
+    protected String registry() {
+    	if (dockerExecutable.endsWith("podman") || Boolean.parseBoolean(env.get(DOCKER_REGISTRY_HOST_ONLY, "false"))) {
```
mmusenbr, line 118: is this podman check ideal? Better than not having it maybe, but it will not detect if the podman-docker wrapper is used.

mmusenbr:
From what I read, this only applies for podman >= 3.3 and <= 4.0. This is fixed in 4.1 in podman.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
rsandell: I don't know anything about podman, so I am not any help here :)
But I guess if anyone comes up with something better it can be enhanced in the future.

rsandell (approved):

PR #166, opened 2025-05-29
[JENKINS-75679] `KeyMaterial` variant passing `VirtualChannel` to `close`

jglick:
Needed for https://github.com/jenkinsci/docker-workflow-plugin/pull/351.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
@@ -29,6 +29,7 @@ import java.net.URL;
 import java.nio.charset.StandardCharsets;
 
 import edu.umd.cs.findbugs.annotations.NonNull;
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import org.apache.commons.lang.StringUtils;
 import org.jenkinsci.plugins.docker.commons.credentials.KeyMaterial;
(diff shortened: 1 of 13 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
```diff
@@ -121,21 +124,38 @@ protected String registry() {
...
+            super(null);
+            assert false : "only deserialized";
         }
 
         @Override
         public void close() throws IOException {
             try {
                 dockerConfig.deleteRecursive();
             } catch (InterruptedException x) {
-                // TODO would better have been thrown from KeyMaterial.close to begin with
```
jglick, line 138: (also addressed with new signature)

src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
```diff
@@ -121,21 +124,38 @@ protected String registry() {
     	return endpoint.toString();
     }
 
-    private static class RegistryKeyMaterial extends KeyMaterial {
+    private static class RegistryKeyMaterial2 extends KeyMaterial2 {
 
-        private final FilePath dockerConfig;
+        private final String dockerConfig;
```
jglick, line 129: The key fix: avoids saving a `FilePathPickle` in `program.dat`.

PR #230, opened 2026-07-20
Remove commons-lang2 usages

gbhat618:
Jenkins core will be dropping commons-lang2, and we would need to remove the usages in all plugins.
Instead of migrating to commons-lang3 - https://github.com/jenkinsci/docker-commons-plugin/pull/174, lets use the jdk apis directly as the usages are small and simple.

### Testing done

* maven compilation works.
* code analysis manual and with AI as simple changes.

### Submitter checklist
- [ ] Link to relevant issues in GitHub or Jira
- [x] Link to relevant pull requests, esp. upstream and downstream changes
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
+++ b/src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
@@ -26,7 +26,6 @@ package org.jenkinsci.plugins.docker.commons.credentials;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.NonNull;
 import hudson.util.FormValidation;
-import org.apache.commons.lang.StringUtils;
 
 import java.util.Arrays;
 import java.util.regex.Pattern;
@@ -59,7 +58,7 @@ public class ImageNameValidator {
      */
     public static @NonNull String[] splitUserAndRepo(@NonNull String userAndRepo) {
         String[] args = new String[4];
-        if (StringUtils.isEmpty(userAndRepo)) {
+        if (userAndRepo.isEmpty()) {
             return args;
         }
         int slashIdx = userAndRepo.lastIndexOf('/');
@@ -108,8 +107,10 @@ public class ImageNameValidator {
             return FormValidation.ok();
         }
         final String[] args = splitUserAndRepo(userAndRepo);
-        if (StringUtils.isBlank(args[0]) && StringUtils.isBlank(args[1]) && StringUtils.isBlank(args[2])
-                && StringUtils.isBlank(args[3])) {
+        if ((args[0] == null || args[0].isBlank())
+                && (args[1] == null || args[1].isBlank())
+                && (args[2] == null || args[2].isBlank())
+                && (args[3] == null || args[3].isBlank())) {
             return FormValidation.error("Bad imageName format: %s", userAndRepo);
         }
         final FormValidation name = validateName(args[1]);
@@ -181,7 +182,7 @@ public class ImageNameValidator {
         if (SKIP) {
             return FormValidation.ok();
         }
-        if (StringUtils.isEmpty(digest)) {
+        if (digest == null || digest.isEmpty()) {
             return FormValidation.ok();
         }
         if (digest.startsWith("@sha256")) { 
@@ -232,7 +233,7 @@ public class ImageNameValidator {
         if (SKIP) {
             return FormValidation.ok();
         }
-        if (StringUtils.isEmpty(tag)) {
+        if (tag == null || tag.isEmpty()) {
             return FormValidation.ok();
         }
         if (tag.length() > 128) {
@@ -280,7 +281,7 @@ public class ImageNameValidator {
         if (SKIP) {
             return FormValidation.ok();
         }
-        if (StringUtils.isEmpty(name)) {
+        if (name == null || name.isEmpty()) {
             return FormValidation.error("Missing name.");
         }
         if (VALID_NAME_COMPONENT.matcher(name).matches()) {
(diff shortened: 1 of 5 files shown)


src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
```diff
@@ -57,9 +56,9 @@ public static boolean skipped() {
      * @return an array where position 0 is the namespace, 1 is the name and 2 is the tag and 3 is the digest.
      *         Any position could be <code>null</code>
      */
-    public static @NonNull String[] splitUserAndRepo(@NonNull String userAndRepo) {
+    public static @NonNull String[] splitUserAndRepo(@CheckForNull String userAndRepo) {
         String[] args = new String[4];
-        if (StringUtils.isEmpty(userAndRepo)) {
+        if (userAndRepo ==null || userAndRepo.isEmpty()) {
```
gbhat618: per the previous `@NonNull`, when there was no null check, test failed as - https://github.com/jenkinsci/docker-commons-plugin/blob/ee120e23d3ac200a3dcad3873f49a043abc015d9/src/test/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidatorTest.java#L53

```
java.lang.NullPointerException: Cannot invoke "String.isEmpty()" because "userAndRepo" is null
	at org.jenkinsci.plugins.docker.commons.credentials.ImageNameValidator.splitUserAndRepo(ImageNameValidator.java:61)
	at org.jenkinsci.plugins.docker.commons.credentials.ImageNameValidator.validateUserAndRepo(ImageNameValidator.java:109)
	at org.jenkinsci.plugins.docker.commons.credentials.ImageNameValidatorTest.test(ImageNameValidatorTest.java:75)
```

By seeing the test, i think it was meant to be `@CheckForNull` instead of `@NonNull` - usage of `StringUtils.isEmpty` didn't flag it before.

Here updating to keep the same behavior instead of updating in the call chain.

gbhat618: ~~wait. now the return value to be `@NonNull` is wrong too...~~

gbhat618: (all good, the reasoning seems correct)

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
```diff
@@ -57,9 +56,9 @@ public static boolean skipped() {
      * @return an array where position 0 is the namespace, 1 is the name and 2 is the tag and 3 is the digest.
      *         Any position could be <code>null</code>
      */
-    public static @NonNull String[] splitUserAndRepo(@NonNull String userAndRepo) {
+    public static @NonNull String[] splitUserAndRepo(@CheckForNull String userAndRepo) {
         String[] args = new String[4];
-        if (StringUtils.isEmpty(userAndRepo)) {
+        if (userAndRepo == null || userAndRepo.isEmpty()) {
```
jglick: It should be `@NonNull`. The only caller is marked non-null! I think you were confused by https://github.com/jenkinsci/docker-commons-plugin/blob/89fe5df5f70c3328a9577bbb872cf254e10f2fdb/src/test/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidatorTest.java#L53 (from #93) which was probably just written thoughtlessly.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
```diff
@@ -181,7 +182,7 @@ public static void checkUserAndRepo(@NonNull String userAndRepo) throws FormVali
         if (SKIP) {
             return FormValidation.ok();
         }
-        if (StringUtils.isEmpty(digest)) {
+        if (digest == null || digest.isEmpty()) {
```
jglick, line 185: Unclear why these methods were `public`. Do not seem to be called from Stapler.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
```diff
@@ -83,7 +82,7 @@ public KeyMaterial2 materialize2() throws IOException, InterruptedException {
         FilePath configJsonPath = FilePath.getHomeDirectory(this.launcher.getChannel()).child(".docker").child(DOCKER_CONFIG_FILENAME);
         if (configJsonPath.exists()) {
             String configJson = configJsonPath.readToString();
-            if (StringUtils.isNotBlank(configJson)) {
+            if (configJson != null && !configJson.isBlank()) {
```
jglick: Could not be null!

jglick (changes requested):
0fbbe9830dc694e757651d72699018f467b2e0e4 resolved incorrectly I think.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
gbhat618: reverted.

src/main/java/org/jenkinsci/plugins/docker/commons/credentials/ImageNameValidator.java
gbhat618: ty. removed the null check, and removed the test case.

src/main/java/org/jenkinsci/plugins/docker/commons/impl/RegistryKeyMaterialFactory.java
gbhat618: noted. ty. removed the null check.

jglick (approved):

